PR #1, opened 2021-05-26
RFC: incentivization-transaction

shelbyd:
RFC for incentivization transactions. Most details will be resolved in implementation.


rfcs/0001-incentivization-transactions.md
```diff
@@ -0,0 +1,66 @@
...
+
+[summary]: #summary
+
+Provide a mechanism to pay users for having a verified identity and continuously adding data to their data host.
+
+# Motivation
+
+[motivation]: #motivation
+
+The protocol is deeply dependant on the network effects of users wanting applications and applications wanting users.
```
juliosantos, line 11: I think "users wanting applications and applications wanting users" would be clearer as "data providers wanting data consumers, and data consumers wanting data providers". "Applications" is a bit context-free here

rfcs/0001-incentivization-transactions.md
```diff
@@ -0,0 +1,66 @@
...
+
+Mint tokens for addresses that publish the following:
+
+1. A unique ID signed by Fractal ID.
+1. A Merkle root containing more data than the previously published one for that Fractal ID.
+
+We will mint tokens with the following constraints:
+
+1. Constant amount per user per day, when below total issuance per day.
+   This ensures users are indifferent towards more users using the system.
```
juliosantos, line 37: I'm not sure I agree with this, because while it helps with not discouraging late adopters (by flooring their upside), it hurts encouraging early ones (by capping it). Additionally, rewarding early adopters outsizedly seem to be in line with the expectations of the community and the practices of other projects.

rfcs/0001-incentivization-transactions.md
```diff
@@ -0,0 +1,66 @@
...
+1. Maximum total issued per day, with minting evenly distributed among IDs.
+
+Additionally, communicate to users that we will retroactively be incentivizing genuine data.
+Users should have these Merkle roots include their genuine browsing data, as they will be paid for that later.
```
juliosantos, line 41: Does this mean "if you provide data at t0 and only hold a signed ID at t1, we'll pay you back the incentives from t0"?

rfcs/0001-incentivization-transactions.md
```diff
@@ -0,0 +1,66 @@
+# Summary
+
+[summary]: #summary
+
+Provide a mechanism to pay users for having a verified identity and continuously adding data to their data host.
```
juliosantos, line 5: I know technically it's all the same, but depending on the audience it might make sense to clarify that "user" might also mean "a user-facing application"

rfcs/0001-incentivization-transactions.md
shelbyd: My concern is that if we issue a constant amount daily divided among users, then users are incentivized to limit the number of others getting these rewards. As it is, we can tune the max amount to be quite high, but not an entire day's worth of minting. We can also tune the amount so that it starts being evenly distributed at N total users, whatever we decide a good N is.

rfcs/0001-incentivization-transactions.md
shelbyd: It means that once we have the infrastructure to analyze data for genuineness and mint based on that, we will then mint somehow including data from this earlier phase of minting. So capturing genuine data now will get you paid more later than if you just started capturing genuine data later.

juliosantos: So a scenario would be that we distribute incentives to those with a signed ID and data, AND promise users that later, when we deploy additional mechanisms to ascertain genuineness, those will apply retroactively?

rfcs/0001-incentivization-transactions.md
juliosantos: That's a valid concern. Setting a max makes sense to me

rfcs/0001-incentivization-transactions.md
shelbyd: I believe so. I think we want to maximize the amount of genuine data that we capture before we have the infrastructure to analyze it. 

It would be irrelevant if we launch the mainnet with the ability to incentivize based on data genuineness. In that case, we don't have this incentivization phase, so no reason to make promises for it.

juliosantos: Sounds good to leave the option open, but let's take care to avoid setting expectations we don't want to break :)

rfcs/0001-incentivization-transactions.md
shelbyd: I don't think that's necessary to include here. I also want to keep the focus on the people using the protocol as much as possible.

rfcs/0001-incentivization-transactions.md
shelbyd: Done.

rfcs/0001-incentivization-transactions.md
shelbyd: Rephrased some.

juliosantos (approved):

PR #2, opened 2021-07-06
Basic SIER codec implementation

shelbyd:
Beginning implementation laying out the general structures we'll use for the future.


support/sier-codec/src/definition_parser.rs
```diff
@@ -0,0 +1,150 @@
...
+    Unresolved(&'i str),
+}
+
+impl<'i> ParsedStruct<'i> {
+    fn compile(self) -> Result<StructDef, Error<'i>> {
+        use std::collections::HashSet;
+
+        let mut seen = HashSet::with_capacity(self.fields.len());
+        for field in &self.fields {
+            dbg!(&field);
```
boymaas, line 36: what are the guidelines concerning dbg! statements in to be merged commits, remove them?

support/sier-codec/src/schema.rs
```diff
@@ -0,0 +1,245 @@
...
+        hasher.update(&self.type_name);
+        for field in &self.fields {
+            hasher.update(&field.name);
+            hasher.update(field.type_.id());
+        }
+
+        let digest = hasher.finalize();
+
+        let bytes: &[u8] = &digest[0..8];
+        bytes.try_into().expect("hash should always be > 8 bytes")
```
boymaas, line 30: this is a panicing function, I remember from our conversations we want to avoid this in the blockchain code, is this lib ever going to be used in the blockchain implementation?

support/sier-codec/src/schema.rs
```diff
@@ -0,0 +1,245 @@
...
+            let struct_b = StructDef {
+                type_name: "Foo".to_string(),
+                fields: vec![FieldDef {
+                    name: "bar".to_string(),
+                    type_: Type::U8,
+                }],
+            };
+
+            assert_eq!(struct_a.id(), struct_b.id());
+        }
```
boymaas, line 243: panic case on id is not covered, maybe because of logical reasons it can never occur?

support/sier-codec/tests/integration.rs
```diff
@@ -0,0 +1,41 @@
+use sier_codec::Parser;
```
boymaas, line 1: maybe a small nuance, as you shared integration tests are a scam, we can rename the "integration test" to "integrated test", this to remind ourselves to use that paradigm (if this is the case for this project) => integrated.rs

support/sier-codec/src/schema.rs
shelbyd: Yes. They should be removed. I missed this one.

support/sier-codec/tests/integration.rs
shelbyd: Good idea. Done.

support/sier-codec/src/schema.rs
shelbyd: The expect should never panic. We know the digest output is always 64 bytes.

I don't see a combination of methods in generic_array (https://docs.rs/generic-array/0.14.4/generic_array/struct.GenericArray.html) and std to get a subarray of fixed size from a larger one. So we just use the easier methods that return Option<[u8; 8]> that we need to unwrap.

support/sier-codec/src/definition_parser.rs
shelbyd: They should be removed. Done.

support/sier-codec/src/schema.rs
shelbyd: This will probably not be used in the blockchain. I'm not aware of a way to get a fixed size array from a larger fixed size array without doing some type of unwrap.

I just found the Split impl (https://docs.rs/generic-array/0.14.4/generic_array/struct.GenericArray.html#impl-Split%3CT%2C%20K%3E) for generic array. I'll see if I can use that to not look like it may panic.

shelbyd: It seems hard/impossible to convert a GenericArray to a fixed size array, even if the types line up. I have an approach that could work to make it not look like it might panic, but I don't think it's quite worth it ATM. Added TODO.

boymaas: one way of solving this could be to check on the byte size and return a Result based on this, I am not deep enough in the context of this function to determine the probability it will panic. If this panic cannot happen (from the structure or usage) this is not an issue, otherwise maybe do this size check.

shelbyd:
There's no possibility of it actually panicing. The method will always return Some(...).

boymaas (approved):

PR #4, opened 2021-07-07
Add a basic data tracking Chrome extension

frm:
Why:

* We want our data host to gather data on the user.

* Converting the existing Chrome extension to a React + Redux extension.
* This allows us to easily synchronise and communicate between the
  content script, background script and extension pages.
* Adding a simple tracker of visited websites

Notes:

* The tracker is keeping everything locally for now.
* We should discuss exactly what data and what format we want to
  discuss.
* We should also discuss how and where to relay the data. Chrome
  extensions are limited to storing 5MBs of data in the local storage.
* We could optionally add the `unlimitedStorage` permission but this is
  __very__ eager and should generally be avoided. Users might also be
  wary of such extensions.

Notes for review:

* All of these files are automatically generated from https://github.com/lxieyang/chrome-extension-boilerplate-react. To review, focus on the following files:
  * `redux/*`
  * `lib/*`
  * `Background/index.js`
  * `Content/index.js`
  * `Popup/*`


shelbyd (dismissed):
If it ever comes up again, please separate changing the boilerplate from adding a feature. That way reviewers can easily LGTM the boilerplate and actually review the feature code. In general, multiple smaller PRs are better.

boymaas (approved):

PR #6, opened 2021-07-08
Parse string fields

shelbyd:
Users can now define a field as a string. The encoding expects a length prefixed sequence of valid utf8 bytes. The length prefix is a varint. See https://developers.google.com/protocol-buffers/docs/encoding#varints.


boymaas (dismissed):

support/sier-codec/src/schema.rs
```diff
@@ -75,16 +75,57 @@ impl FieldDef {
...
+    IResult,
+};
+
+fn length_prefixed(b: &[u8]) -> IResult<&[u8], &[u8]> {
+    let (b, len) = var_int(b)?;
+    take(len)(b)
+}
+
+fn var_int(b: &[u8]) -> IResult<&[u8], usize> {
+    let (new_b, int_bytes) = recognize(pair(take_while(|b| b & 128 > 0), take(1usize)))(b)?;
```
frm, line 102: Just to make sure I get it right: our strings are length prefixed, var int at the header (128 bits). We're shifting that byte by byte until we extract the numeric length value and the remainder which is the string. We then extract that length from the buffer.

Is this used anywhere else than strings?

support/sier-codec/src/schema.rs
```diff
@@ -243,4 +285,129 @@ mod tests {
...
+            }
+
+            #[test]
+            fn non_utf8() {
+                let field = FieldDef {
+                    name: "foo".to_string(),
+                    type_: Type::String,
+                };
+
+                assert!(field.parse(&[4, 0, 159, 146, 150]).is_err());
```
frm, line 330: Normally I don't really care about error types in applications, but since this is library code and we set a very specific error at https://github.com/trustfractal/protocol/pull/6/files#diff-d2c7017b3ce07f0414d04f03a7ec686c452d012307c09db316c9b176a7a4cb5dR57, can we assert that the error is the correct one and not a different type of error we might be causing?

For now, it's fairly obvious what the output would be but as the codebase grows, it might be helpful to keep track of that.

support/sier-codec/src/schema.rs
```diff
@@ -243,4 +285,129 @@ mod tests {
...
+            }
+
+            #[test]
+            fn fewer_bytes_than_size_says() {
+                let field = FieldDef {
+                    name: "foo".to_string(),
+                    type_: Type::String,
+                };
+
+                assert!(field.parse(&[3, 65, 66]).is_err());
```
frm, line 363: Same as the previous comment, but for the `TooFewBytes` error.

frm (changes requested):

support/sier-codec/src/schema.rs
shelbyd: The "varint" representation will be used in a lot of places. It's used for any length for the codec (like length of strings or length of lists). It can also be the representation of numbers that are disproportionately distributed near 0. Varints are the default for numbers in protobuf. SCALE also has a similar approach to reducing the byte-size of small-ish numbers.

support/sier-codec/src/schema.rs
shelbyd: Done.

support/sier-codec/src/schema.rs
shelbyd: Done.

boymaas (approved):

frm (approved):

PR #11, opened 2021-07-13
Add basic ability to construct codec objects in memory

shelbyd:


boymaas (dismissed):

support/sier-codec/src/builder.rs
```diff
@@ -0,0 +1,160 @@
...
+
+use std::collections::HashMap;
+
+pub struct Builder<'s> {
+    struct_def: &'s StructDef,
+    field_values: HashMap<String, Value>,
+}
+
+impl<'s> Builder<'s> {
+    pub(crate) fn new(struct_def: &'s StructDef) -> Self {
```
frm, line 11: Why the use of `pub(crate)` here?

support/sier-codec/src/builder.rs
```diff
@@ -0,0 +1,160 @@
...
+        self
+    }
+
+    pub fn try_build(mut self) -> Result<Object<'s>, BuildError> {
+        let values = self
+            .struct_def
+            .fields
+            .iter()
+            .map(|field| self.value_for_field(field))
+            .collect::<Result<Vec<_>, _>>()?;
```
frm, line 30: This could be a single call to `std::iter::fold`, couldn't it?

support/sier-codec/src/builder.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    pub fn set(mut self, field_name: &str, value: impl Into<Value>) -> Self {
+        self.field_values
+            .insert(field_name.to_string(), value.into());
+        self
+    }
+
+    pub fn try_build(mut self) -> Result<Object<'s>, BuildError> {
```
frm, line 24: Are we not adding the remaining two `build` methods?

support/sier-codec/src/builder.rs
```diff
@@ -0,0 +1,160 @@
...
+                name: "bar".to_string(),
+                type_: Type::U8,
+            }],
+        };
+
+        let obj = def.builder().try_build();
+
+        assert_eq!(
+            obj.err(),
+            Some(BuildError::MissingField("Foo.bar".to_string()))
```
frm, line 99: 👍

frm (commented):
LGTM but don't forget to check my comments 🙏

support/sier-codec/src/builder.rs
shelbyd: Because this method is used in the StructDef#builder method. I don't want it to be full public, but it can't be private.

support/sier-codec/src/builder.rs
shelbyd: I don't think so. I think this is the best way to write it as it is.

Fold is used when you need to accumulate a value across everything in an iterator. Each of these values is independent of each-other, so I don't think fold is right/necessary.

support/sier-codec/src/builder.rs
shelbyd: I don't want to do it right now. I just want to get to the point that you can start constructing and storing these. I don't think it will be a big effort to add them later.

frm (approved):

boymaas (approved):

PR #12, opened 2021-07-13
Implement Value methods with a macro

shelbyd:
Any ideas about how to do the "as_something" methods with the macro? I'm unsure how to specify which need to be by reference and which can be deref'd.

### DIFF
--- a/support/sier-codec/src/object.rs
+++ b/support/sier-codec/src/object.rs
@@ -27,24 +27,43 @@ impl<'s> Index<&'_ str> for Object<'s> {
     }
 }
 
-#[derive(Debug, PartialEq, Eq)]
-pub enum Value {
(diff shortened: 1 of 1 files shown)


boymaas (dismissed):

frm (approved):

boymaas:
> Any ideas about how to do the "as_something" methods with the macro? I'm unsure how to specify which need to be by reference and which can be deref'd.

It is a bit unclear to me what you exactly mean, do you have a pointer to an example (already implemented) in another codebase of the behaviour you want?

shelbyd:
I don't have an example, and I'm not sure how I feel about the structure of this module ATM in general. I'm toying around with a few different ideas for refactoring it. But those will be in a future PR.

PR #13, opened 2021-07-13
Allow the user to input a Fractal ID

frm:
Why:

* We want the users to be able to save and submit their own Fractal IDs
  so we can submit them to the blockchain.

* Adding minimal code to save said ID to the current state.

https://user-images.githubusercontent.com/4325027/125503965-7a14bcd0-e834-4c71-8606-28245a247650.mov

### DIFF
--- a/data_host/chrome_extension/src/pages/Popup/Popup.jsx
+++ b/data_host/chrome_extension/src/pages/Popup/Popup.jsx
@@ -1,7 +1,9 @@
 import React from 'react';
-import { useSelector } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
+import { Form, Field } from 'react-final-form';
 
-import { getWebpages } from '../../redux/selectors';
+import { getWebpages, getFractalData } from '../../redux/selectors';
+import { addFractalID } from '../../redux/actions';
 
 import './Popup.css';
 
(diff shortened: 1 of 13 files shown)


data_host/chrome_extension/src/pages/Popup/Popup.jsx
```diff
@@ -27,18 +26,86 @@ const renderWebpage = (hostname, paths, key) => {
...
+  await dispatch(addFractalID(fractalID));
+  form.reset();
+};
+
+const validate = (data) => {
+  switch (true) {
+    case Object.keys(data).length === 0:
+      return {};
+    case !data.fractalID.startsWith('0x'):
+      return { fractalID: 'Must be a valid message' };
```
shelbyd, line 52: This is one my big pet peeves. Anything that says "do it right" without saying what "right" is.

Instead, say something like "Expected hex-encoded byte sequence beginning with '0x'".

shelbyd (dismissed):

data_host/chrome_extension/src/pages/Popup/Popup.jsx
frm: I agree but honestly, I didn't give it much thought since it was just an example validation to ensure it would work. Ideally, I'd check Fractal's signature here. I'll replace the text.

shelbyd (approved):

PR #17, opened 2021-07-13
Add simple method to serialize object

shelbyd:
Next PR will add quickcheck testing that arbitrary objects can be serialized and deserialized with the resulting objects matching.

### DIFF
--- a/support/sier-codec/src/object.rs
+++ b/support/sier-codec/src/object.rs
@@ -11,6 +11,15 @@ impl Object<'_> {
     pub fn schema(&self) -> &StructDef {
         self.schema
     }
+
+    pub fn serialize(&self) -> Vec<u8> {
+        let mut result = Vec::new();
+        result.extend(self.schema.id().iter().cloned());
+        for value in &self.values {
+            result.extend(value.serialize());
+        }
+        result
+    }
 }
 
 impl<'s> Index<&'_ str> for Object<'s> {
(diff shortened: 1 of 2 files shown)


boymaas (approved):

support/sier-codec/src/object.rs
```diff
@@ -11,6 +11,15 @@ impl Object<'_> {
     pub fn schema(&self) -> &StructDef {
         self.schema
     }
+
+    pub fn serialize(&self) -> Vec<u8> {
+        let mut result = Vec::new();
+        result.extend(self.schema.id().iter().cloned());
+        for value in &self.values {
+            result.extend(value.serialize());
```
frm, line 19: Out of curiosity, why the use of `std::vec::Vec::extend` rather than `push`?

support/sier-codec/tests/integrated.rs
```diff
@@ -57,6 +57,30 @@ fn construct_numbers() {
...
+        .try_build()
+        .unwrap();
+
+    let encoded = message.serialize();
+    let expected = def
+        .id()
+        .iter()
+        .chain(&[42, 0, 0, 0, 0, 0, 0, 0, 43, 0, 0, 0])
+        .cloned()
+        .collect::<Vec<_>>();
```
frm, line 79: I was thinking about this regarding the testing philosophy. Is there any way we can avoid computing the expected value? The reasoning is that if we're computing the expected, there is computational logic involved and that can be faulty and hideaway bugs. It also helps with clarity for the reader to know exactly the expected value. However, it's not always practical to give out a hard-coded value. So if it's not possible, this should be fine, but, in general, I'd prefer to have the expectation hard-coded and easier on the eye. What are your thoughts?

frm (dismissed):

support/sier-codec/src/object.rs
shelbyd: Extend is basically push for when you want to push many things that are already in an iterator.

Also, extend can be more efficient when the size of the iterator you're extending with is known. Basically because it can allocate once if necessary, where push may allocate and move the contents multiple times.

support/sier-codec/tests/integrated.rs
shelbyd: The main reason we're doing this is to specify it starts with the ID. I think it'll be better to check different regions of the output as equal.

```
assert_eq!(&result[0..8], struct_def.id());
assert_eq!(&result[8..], &[42, 0, 0, 0]);
```

I'm not gonna do that in this PR. I'll probably do it next time I'm somewhere that is doing this type of work.

support/sier-codec/src/object.rs
shelbyd: Extend is push when you want to push all the values in an iterator.

Also, extend can be more efficient if the size of the iterator is known. Then the Vector only has to allocate at most once, when with a sequence of pushes, it may have to reallocate many times.

frm (approved):

PR #20, opened 2021-07-15
Visually refactor the extension

frm:
Why:

* The codebase was messy with a lot of random styles and cluttered code.

* Adding `styled-components`.
* Refactoring most of the shared code into separate components.
* Applying consistent styles, sizes and spacings across components.


boymaas (approved):

data_host/chrome_extension/src/pages/Popup/components/Form/index.tsx
```diff
@@ -3,6 +3,10 @@ import { useDispatch } from 'react-redux';
 import { Form as FinalForm, Field, FormRenderProps } from 'react-final-form';
 import { FormApi } from 'final-form';
 
+import Error from '../../../../components/Error';
```
shelbyd, line 6: I'm generally not a fan of importing from relative paths for parent directories. If we move this file and not the other one, this import needs to change. it's also really unclear where this is coming from. 

Are we able to import from src/components/Spacing instead? If not, can we change some config so we can?

shelbyd (changes requested):
nit: I believe the term "Refactor" should only be used when only changing the internal structure of code and not the user-facing behavior. Just know if I ever call something refactoring (and I will a lot) I mean explicitly to not change anything out the external facing behavior.

Perhaps this could be considered a visual refactoring insofar as changing the visual design without changing functionality. Consider changing the title, nbd if you want to keep as is.

data_host/chrome_extension/src/pages/Popup/components/Form/index.tsx
frm: Just pushed configuration including path aliases 👍

frm:
Closing due to changing focus to [Fractal Wallet](https://github.com/trustfractal/wallet)

PR #24, opened 2021-07-16
Boymaas cicd docker build

boymaas:
Docker build and push phase of the CD. After going through the route of a custom rust-build container, and a multistage Dockerfile with caching using cargo chef. Due to infrastructure issues and timeouts I brought it back to the most robust solution. Removed the Dockerfiles from the dir and generate them in the .circleci/config.yml. 

This uses a combination of workspaces and coupling the ubuntu version used in circleci/rust base image (ubuntu:focal). Currently we are deploying  a *debug* version, which has been tested. Once the *release* build and test builds succeed we can push the release version.


shelbyd:
If this is ready to be reviewed, please explicitly add reviewers in the reviewers section on the sidebar.

.circleci/config.yml
```diff
@@ -58,6 +58,10 @@ jobs:
       - run:
           name: Run all tests
           command: cargo test
+      - persist_to_workspace:
+          root: .
+          paths:
+            - target/debug/node
```
frm, line 64: This will persist across the jobs and then be used in the deploy step. But do we want it to? Aren't there extra artifacts included for testing or special flags for production compilation?

.circleci/config.yml
```diff
@@ -143,3 +202,11 @@ workflows:
...
             tags:
               only: /.*/
+      - deploy_node:
+          requires:
+            - build-and-test
+          filters:
+            branches:
+              only:
+                - master
+                - boymaas-cicd-docker-build
```
frm, line 212: Don't forget to remove this before merging

frm (changes requested):

.circleci/config.yml
```diff
@@ -122,6 +129,58 @@ jobs:
...
+            cat \<<EOS >Dockerfile
+            FROM ubuntu:focal
+
+            WORKDIR /app
+
+            COPY target/debug/node /app
+            ENTRYPOINT ["/app/node"]
+            EOS
+
+            # comment needs to be here
```
shelbyd, line 165: Can you say why this comment needs to be here?

.circleci/config.yml
```diff
@@ -122,6 +129,58 @@ jobs:
...
+
+      - run:
+          name: Setup custom environment variables
+          command: |
+            export BUILD_ID="${CIRCLE_SHA1:0:7}_$(date +'%Y%m%d_%H%M')"
+            export IMAGE="boymaas/nodefcl:$BUILD_ID"
```
shelbyd, line 171: Can we have a TODO somewhere to make a Docker account for fractal and putting the base image there?

shelbyd (changes requested):

.circleci/config.yml
boymaas: As discussed briefly in the retrospective, the reason we chose to use persisting (now debug mode) is the other option is to do a full Dockerfile build, which takes around 50 minutes. Probably because the docker building (as used on circleci) does a single core build, or is magnitudes slower.

.circleci/config.yml
boymaas: Yes we can. Will add it.

.circleci/config.yml
boymaas: Done

shelbyd (approved):

PR #27, opened 2021-07-25
Convert identity registration to use an extrinsic directly.

shelbyd:
I believe setting the weight to 0 will make this not require transaction fees from the Fractal server.

### DIFF
--- a/blockchain/pallets/minting/src/lib.rs
+++ b/blockchain/pallets/minting/src/lib.rs
@@ -14,7 +14,7 @@ pub mod pallet {
     use frame_system::pallet_prelude::*;
 
     use blake2::Blake2b;
-    use codec::{Decode, Encode};
+
     use core::convert::TryInto;
     use frame_support::{
         dispatch::Vec,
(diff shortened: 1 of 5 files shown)


blockchain/pallets/minting/src/lib.rs
```diff
@@ -150,46 +107,37 @@ pub mod pallet {
...
         MustSpecifyFractalIdWithMultipleIds,
         FractalIdNotRegisteredToAccount,
+        MustBeFractal,
     }
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
-        /// Register the provided Fractal ID to the origin account.
-        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(2, 2))]
+        #[pallet::weight(0)]
```
frm, line 119: Out of curiosity: could this be computed in terms of who the origin is?

frm (approved):
LGTM

blockchain/pallets/minting/src/lib.rs
shelbyd: Yes. By returning DispatchResultWithPostInfo I think. https://docs.rs/frame-support/2.0.0-alpha.6/frame_support/dispatch/type.DispatchResultWithPostInfo.html

PR #28, opened 2021-08-05
Set both minting extrinsics to be free

shelbyd:


shelbyd:
@boymaas We'll need to upgrade the testnet machines once this is merged.

boymaas (approved):

PR #30, opened 2021-08-12
Minting config tweaks

boymaas:
Update the minting configuration to mint 80k total in one minting period, and 1.5 units max per user. Update minting period ot one day or 14400 blocks, assuming 6000ms per block


blockchain/runtime/src/lib.rs
```diff
@@ -266,9 +266,9 @@ impl pallet_sudo::Config for Runtime {
 }
 
 parameter_types! {
-    pub const MaxRewardPerUser: Balance = 1 << 60;
-    pub const MaxMintPerPeriod: Balance = 1 << 64;
-    pub const MintEveryNBlocks: BlockNumber = 10;
+    pub const MaxRewardPerUser: Balance = 1.5e18 as Balance;
```
shelbyd, line 269: Does rust guarantee that `1.5e18 as u128` actually produces 1,500,000,000,000,000,000? It seems to do that from my tests, but I'm not sure about guarantees...

I think I'd prefer something like `1_500 * MILLI_UNIT` for this and `80_000 * UNIT` for max mint.

shelbyd (approved):

PR #32, opened 2021-08-12
Devnet deployment

boymaas:
Calls the deployment script on the devnet server completing the development  CD

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,6 +183,12 @@ jobs:
           name: Push docker image
           command: |
             docker push $IMAGE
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -212,4 +221,4 @@ workflows:
             branches:
               only:
                 - master
-                - boymaas-cicd-docker-build
+                - bmaas/devnet-deploy
```
shelbyd, line 224: I think we should remove this line? Seems like something only here for testing.

shelbyd (approved):

PR #34, opened 2021-08-17
First call to register_for_minting is free

boymaas:


blockchain/pallets/minting/src/lib.rs
```diff
@@ -152,13 +152,13 @@ pub mod pallet {
         #[pallet::weight((
             10_000 + T::DbWeight::get().reads_writes(2, 2),
             DispatchClass::Normal,
-            Pays::No
+            Pays::Yes
```
shelbyd, line 155: I don't think the Substrate node will accept this transaction if we have this as Pays::Yes.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -284,6 +284,8 @@ mod register_identity {
 
     #[cfg(test)]
     mod extension_proofs {
```
shelbyd, line 286: This shouldn't be in the extension_proofs module. Probably goes in the root test module next to the other tests about minting registration.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -338,5 +340,39 @@ mod register_identity {
                 );
             });
         }
+
+        #[test]
+        fn register_for_minting_is_free_for_first_call() {
```
shelbyd, line 345: I'd suggest looking at some of the other tests. We have functions already for registering for minting with merkle trees and other utilities for the relevant steps.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -338,5 +340,39 @@ mod register_identity {
...
+            new_test_ext().execute_with(|| {
+                register_id_account(42, 1);
+
+                // first call is free, actual_weight set to 0
+                let tree_0 = gen_tree(&["a", "b"]);
+                assert_eq!(
+                    FractalMinting::register_for_minting(Origin::signed(1), Some(42), tree_0),
+                    Ok(PostDispatchInfo {
+                        actual_weight: Some(0),
+                        pays_fee: Pays::Yes
```
shelbyd, line 361: Does it not work if the first PostDispatchInfo specifies Pays::No and the second one Pays::Yes?

blockchain/pallets/minting/src/tests.rs
```diff
@@ -338,5 +340,39 @@ mod register_identity {
...
+
+            new_test_ext().execute_with(|| {
+                register_id_account(42, 1);
+
+                // first call is free, actual_weight set to 0
+                let tree_0 = gen_tree(&["a", "b"]);
+                assert_eq!(
+                    FractalMinting::register_for_minting(Origin::signed(1), Some(42), tree_0),
+                    Ok(PostDispatchInfo {
+                        actual_weight: Some(0),
```
shelbyd, line 360: I think we still want the first transaction to have weight, otherwise we might be able to overload blocks.

blockchain/pallets/minting/src/lib.rs
```diff
@@ -190,9 +190,18 @@ pub mod pallet {
...
+            // entry in NextMintingRewards. This entry will be removed when
+            // minting rewards get dispersed in the on_finalize block. Not present
+            // indicates its the first call.
+            if NextMintingRewards::<T>::try_get(id).is_ok() {
+                // None indicates use pallet::weight annotation
+                Ok(None.into())
+            } else {
+                NextMintingRewards::<T>::insert(id, who);
+                Ok(Some(0).into())
+            }
```
shelbyd, line 204: With this logic, I'd expect the following test to fail:

register_identity
register_for_minting

run_to_minting
let dispatch_info = register_for_minting
assert!(dispatch_info.pays_fee == Pays::No)

blockchain/pallets/minting/src/lib.rs
```diff
@@ -190,9 +190,18 @@ pub mod pallet {
...
-            Ok(())
+            // Only first call is free, determined by presence of
+            // entry in NextMintingRewards. This entry will be removed when
+            // minting rewards get dispersed in the on_finalize block. Not present
+            // indicates its the first call.
+            if NextMintingRewards::<T>::try_get(id).is_ok() {
+                // None indicates use pallet::weight annotation
+                Ok(None.into())
+            } else {
+                NextMintingRewards::<T>::insert(id, who);
```
shelbyd, line 202: I think if users are already registered for minting and they want to register again for the same period, we should allow it. If only for eventually using the merkle_tree for proving data existence.

shelbyd (changes requested):

blockchain/pallets/minting/src/lib.rs
boymaas: Interesting, I've added the test and you are right, it does not pass. 

```
for (id, account) in recipients { ‣id: &u64 ‣account: &<T as Config>::AccountId
   T::Currency::deposit_creating(account, actual_mint_per_user);
   NextMintingRewards::<T>::remove(id);
}
```

my assumption was that the `NextMintingRewards::<T>::remove(id)` in the `on_finalize` would remove the entry in `NextMintingRewards`. What am I missing here?

blockchain/pallets/minting/src/tests.rs
boymaas: are you referring to the `gen_tree` I have added.?I could not find a utility function which generate a new `MerkleTree` that extends the earlier one.

blockchain/pallets/minting/src/lib.rs
boymaas: Interesting, just checking the returns from this pallet test does not ensure it will be accepted by the node, the test runtime differs as it the test runtime does not go through the  System pallet and the Transaction Payment pallet. Makes sense, as such Pays::No also excludes transaction costs of the intrinsic. I was not aware of these details.

blockchain/pallets/minting/src/lib.rs
boymaas: This is resolved after our conversation, the idea is that only the first call ever should be free, my understanding was that the first call in each "epoch" should be free.

blockchain/pallets/minting/src/tests.rs
boymaas: It has been rewritten to use the Pays::Yes and Pays::No to exclude fee, this automatically puts a weight on the Pays::No transaction as such the overloading  blocks should be resolved.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -282,6 +287,63 @@ mod register_identity {
...
     }
 
+    #[test]
+    fn first_call_to_register_for_minting_is_free() {
+        use frame_support::pallet_prelude::Pays;
+
+        new_test_ext().execute_with(|| {
+            register_id_account(42, 1);
+
+            // first call is free, actual_weight set to 0
```
shelbyd, line 297: This is a stale comment. This is one of the stronger reasons to not add comments to code. For unit tests, the tests should describe what is happening. For production code, when I used to add comments, I now extract functions or variables so I can give something a name.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -282,6 +287,63 @@ mod register_identity {
...
+            // second is charged, actual_weith None indicates annotation
+            // weight is applied
+            let tree_1 = gen_tree(&["a", "b", "c"]);
+            assert_eq!(
+                FractalMinting::register_for_minting(Origin::signed(1), Some(42), tree_1),
+                Ok(PostDispatchInfo {
+                    actual_weight: None,
+                    pays_fee: Pays::Yes
+                })
+            );
```
shelbyd, line 316: I would add this as a separate test. The description of this test says the first minting is free, and this doesn't exactly relate to that. If you're doing work after assertions in a test, that's usually a sign you should have two tests.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -282,6 +287,63 @@ mod register_identity {
...
+        });
+    }
+
+    #[test]
+    fn first_call_to_register_for_minting_is_free_logic() {
```
shelbyd, line 321: I don't know what this test is supposed to be testing from this name.

blockchain/pallets/minting/src/tests.rs
```diff
@@ -282,6 +287,63 @@ mod register_identity {
...
+
+            // first call is free, actual_weight set to 0
+            let tree_0 = gen_tree(&["a", "b"]);
+            assert_eq!(
+                FractalMinting::register_for_minting(Origin::signed(1), Some(42), tree_0),
+                Ok(PostDispatchInfo {
+                    actual_weight: None,
+                    pays_fee: Pays::No
+                })
+            );
```
shelbyd, line 335: Is there a reason you're not using the `register_for_minting_dataset` function in these tests?

blockchain/pallets/minting/src/lib.rs
```diff
@@ -192,7 +196,16 @@ pub mod pallet {
             IdDatasets::<T>::insert(id, extension_proof);
             NextMintingRewards::<T>::insert(id, who);
 
-            Ok(())
+            // Only first call is free, determined by not having an
+            // entry in .CalledRegisterForMinting. All subsequent calls
+            // will be Pays::Yes
+            if CalledRegisterForMinting::<T>::try_get(id).is_ok() {
```
shelbyd, line 202: I don't think we need a new storage for this. We can probably use existence in the IdDatasets storage.

shelbyd, line 202: I think with this logic, I would need to pay a fee if I did the following:

register_identity(address1)
register_for_minting(address1)
register_identity(address2)
register_for_minting(address2)

I wouldn't be able to bootstrap address2 with the logic as is. Bootstrapping an address is the whole reason to not require a fee.

shelbyd (changes requested):

blockchain/pallets/minting/src/tests.rs
boymaas: I agree, this is now covered in second test I created from previous remark

blockchain/pallets/minting/src/tests.rs
boymaas: I agree, I have split this up into 2 tests, where both only have on assert at the end

blockchain/pallets/minting/src/tests.rs
boymaas: now I understand your previous remark, I was looking for a gen_tree remark. Will rewrite it.

blockchain/pallets/minting/src/tests.rs
boymaas: I agree, removed the comments

blockchain/pallets/minting/src/lib.rs
boymaas: I agree, this will also cover the remove when the address changes on a register_identity if I am correct.

boymaas: Agree, this was misunderstood, I implemented only once in an absolute sense. Wat not aware of the requirement of bootstrapping a second address on the same fractal_id

blockchain/pallets/minting/src/tests.rs
```diff
@@ -282,6 +287,65 @@ mod register_identity {
...
+
+            let tree_0 = gen_tree(&["a", "b"]);
+            assert_eq!(
+                FractalMinting::register_for_minting(Origin::signed(1), Some(42), tree_0),
+                Ok(PostDispatchInfo {
+                    actual_weight: None,
+                    pays_fee: Pays::No
+                })
+            );
+        });
```
shelbyd, line 305: I think it would be better to reuse the existing `register_for_minting_dataset` function. You can change that function to return the PostDispatchInfo so you can check `Pays`.

Then these lines would be simply:

```
let post = register_for_minting_dataset(1, ["a", "b"]);
assert_eq!(post.pays_fee, Pays::No);
assert_eq!(post.actual_weight, None);
```

blockchain/pallets/minting/src/lib.rs
shelbyd: Does Substrate accept the transaction if the static is Pays::Yes? I don't think it would for an account with no balance. I would expect it to work in a unit test, but not when running a local node.

It should be fine to statically specify Pays::No and override with Pays::Yes for the relevant cases.

blockchain/pallets/minting/src/lib.rs
```diff
@@ -182,17 +182,26 @@ pub mod pallet {
...
                 );
             }
 
             IdDatasets::<T>::insert(id, extension_proof);
             NextMintingRewards::<T>::insert(id, who);
 
-            Ok(())
+            // Only first call is free, determined by not having an
+            // entry in .CalledRegisterForMinting. All subsequent calls
+            // will be Pays::Yes
```
shelbyd, line 198: The tests explain why we have this code. I strongly think we should avoid comments in code. 

The only reason to include code comments is to explain why we're doing something in way that isn't the expected way.  I've only left code comments in two instances, one when I was calling `eval` in JavaScript, which is a potentially dangerous function, and the other when I was mutating the internals of some Java class because it gave the interface I wanted. Both of the comments were about why I was doing that instead of any alternative.

When I think that code is unclear in what it is doing or what it is for, I make a function with a name that clarifies the code. I don't think this code warrants an explanatory function, but if you do, I'm not opposed (just make sure it doesn't accidentally get exposed to users through some substrate magic).

blockchain/pallets/minting/src/lib.rs
```diff
@@ -182,17 +182,26 @@ pub mod pallet {
...
+            if id_datasets_entry.is_some() {
+                // None indicates use pallet::weight annotation
```
shelbyd, line 200: Same as above about code comments.

shelbyd (changes requested):

blockchain/pallets/minting/src/lib.rs
boymaas: Yes makes sense, I will follow the coding style you prefer. Personally I comment a lot in code.

blockchain/pallets/minting/src/tests.rs
boymaas: Agree, changed

PR #36, opened 2021-08-18
Fix amount of minting reward

shelbyd:
We minted about 1,000,000 times too much for the first minting.


shelbyd:
@boymaas Please merge and update the testnet runtime with this. Please DO NOT reset the state of the testnet yet. We will be resetting the state of the testnet once it will be seamless for users.

boymaas (approved):

PR #39, opened 2021-09-01
Add sql script for calculating user's minting rewards

shelbyd:


scripts/download_chain.sh
```diff
@@ -0,0 +1,7 @@
+#! /usr/bin/bash
+
+set -euxo pipefail
+
+sub2sql --chain=wss://889e30bc.testnet.fractalprotocol.com \
```
juliosantos, line 5: Unsure if ready yet, but wouldn't `nodes.testnet.fractalprotocol.com` be a better host to connect to?

juliosantos (approved):

scripts/minting_rewards.sql
```diff
@@ -0,0 +1,26 @@
+WITH
+  minting_registrations AS (
+    SELECT
+      number / 14400 AS period,
```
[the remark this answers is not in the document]
shelbyd: It is the number of blocks for minting (which is 24 * 60 * 60 / 6).

scripts/download_chain.sh
shelbyd: Done.

PR #40, opened 2021-09-02
Add sql script for detecting thrashing registrations

shelbyd:


juliosantos:
If the point of this is to count, lgtm. Why 6 for problematic?

shelbyd:
Mostly to count and keep around so we don't have to figure it out again. These are just utilities.

6 Is arbitrary, I don't have much reason to pick anything specific.

If looks good, please "Approve" in GitHub's UI. Should be a review button in the top right.

juliosantos (approved):

PR #46, opened 2021-09-07
Allow multiple addresses attached to a FractalID at once

shelbyd:


melatron (approved):

boymaas (approved):

shelbyd:
@boymaas Please don't merge code from a story you don't own (unless it's an emergency).

This wasn't ready to be merged, and if it were deployed, it would break minting registration for all users. I was waiting for the rollout of a new wallet version that is compatible with the old and new runtime.

PR #47, opened 2021-09-08
Allows structs in SIER

melatron:
After all TODOs are implemented and more tests added the PR would be ready for review.

### DIFF
--- a/support/sier-codec/src/builder.rs
+++ b/support/sier-codec/src/builder.rs
@@ -62,7 +62,7 @@ pub enum BuildError {
     IncorrectType {
         field: String,
         expected: Type,
-        got: Type,
+        got: Type<String>,
     },
 }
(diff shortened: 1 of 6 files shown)


shelbyd:
I wasn't sure how to approach the problems you were having, so I pulled the branch and tried a few things. Once I found something that worked, I had most of the code implemented, so I just finished everything else. Feels kinda rude in retrospect, 😬. Feel free to ignore my changes and implement this yourself. 

I think we need the Arcs since we'd otherwise have to reference an object in the Parser, which would require Types to have a 'self ref, which isn't supported in Rust right now.

shelbyd (approved):

shelbyd:
@boymaas @frm I believe Antoni is waiting for one of you to review.

PR #48, opened 2021-09-13
Add script for viewing daily counts of failed extrinsics

shelbyd:


juliosantos (approved):
Might be useful to add `method` and then `GROUP BY 1, 2` instead (although so far I've only seen failed minting registrations)

shelbyd:
These scripts are mostly meant to be a starting point / reference for whatever one needs to do.

PR #52, opened 2021-09-23
Batch transactions

melatron:


melatron:
I think that for the `construct_runtime!` it would be best to jump by 10 for each section(System, Monetary...), so that we can add pallets in a structured way in the future.
@shelbyd what do you think?

support/dev-tools/batch-transactions/index.js
```diff
@@ -0,0 +1,43 @@
...
+        .signAndSend(signer, ({ status }) => {
+            if (status.isInBlock) {
+                console.log(`included in ${status.asInBlock}`);
+            }
+        });
+    console.log(result)
+    return result;
+}
+
+main().catch(console.error);
```
shelbyd, line 43: Is this a file we want to commit to the repo? It seems to only be for your local testing of including the utility pallet.

I'd want to include executable examples for the better polka-dot api we make, but this isn't that.

shelbyd (changes requested):

blockchain/runtime/src/lib.rs
```diff
@@ -295,15 +301,23 @@ construct_runtime!(
...
+        // Monetary
+        Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>} = 5,
+        TransactionPayment: pallet_transaction_payment::{Pallet, Storage} = 6,
+
+        // Handy utilities.
+        Utility: pallet_utility::{Pallet, Call, Event} = 7,
+        Sudo: pallet_sudo::{Pallet, Call, Config<T>, Storage, Event<T>} = 8,
+
+        // Fractal pallets
+        FractalMinting: fractal_minting::{Pallet, Call, Storage, Config<T>, Event<T>} = 9,
```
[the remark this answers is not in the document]
shelbyd: I wouldn't jump each system by any amount. This feels a lot like Protocol Buffer tags, where the specific values don't really matter. I think you'll still have to look up the pallet the specific index corresponds to anyway. I don't think quickly knowing that a pallet is Monetary related helps too much.

The only thing to watch out for is to make sure that if we remove a pallet, we mark the ID for that pallet as reserved so future adds don't try to use the same value.

support/dev-tools/polkadot-utils/src/settings.ts
```diff
@@ -0,0 +1,8 @@
+export default {
+    nodeAddress: "ws://127.0.0.1:9944",
+    types: {
+        FractalId: "u64",
+        MerkleTree: "Raw"
+    },
```
shelbyd, line 6: Are you able to include the JSON file at blockchain/types.json? That will be the source of truth for our types.

support/dev-tools/polkadot-utils/src/lib/TxnBatcher.test.ts
```diff
@@ -0,0 +1,59 @@
+import { TxnBatcher } from './TxnBatcher';
+import { ApiPromise, WsProvider } from '@polkadot/api';
+import { Keyring } from '@polkadot/keyring';
+import settings from '../settings';
+
+async function createPromiseApi(nodeAddress: string, types: any) {
+    const provider = new WsProvider(nodeAddress);
+    const api = await new ApiPromise({ provider, types });
+    await api.isReady;
+    return api;
```
shelbyd, line 10: IIUC, this requires having a local node running for these tests to pass? That's not a good idea. Tests should start up, execute, and finish fully in isolation.

support/dev-tools/polkadot-utils/src/lib/TxnBatcher.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        const signer = keyring.createFromUri('//Alice')
+        const address = keyring.createFromUri('//Bob').address
+        batcher.send(api.tx.fractalMinting.registerIdentity(fractalId, address), signer)
+        batcher.send(api.tx.fractalMinting.registerIdentity(fractalId, address), signer)
+        batcher.send(api.tx.fractalMinting.registerIdentity(fractalId, address), signer)
+        batcher.send(api.tx.fractalMinting.registerIdentity(fractalId, address), signer)
+
+        expect(spy).toBeCalledTimes(2)
+      });
+});
```
shelbyd, line 59: A critical behavior of this class is that if two of them are competing to issue transactions, each transaction sent to one of them eventually gets submitted. Right now, we only try to submit the transaction once and drop it if it fails.

support/dev-tools/polkadot-utils/src/lib/TxnBatcher.ts
```diff
@@ -0,0 +1,40 @@
+export class TxnBatcher {
+    private isInSend = false;
+    private queued = new Map();
+
+    constructor(private readonly api: any) {
+        this.api = api;
+    }
+
+    public async send(txn: any, signer: any) {
```
shelbyd, line 9: This doesn't allow me to wait on this transaction being submitted. A user of this class would do something like the following:

```ts
const hash = await batcher.send(txn, signer);
console.log('Transaction submitted');
return {txn: hash};
```

PR #54, opened 2021-09-30
Seed bridge account with 5,000,000 units

shelbyd:


juliosantos (approved):
Let's make sure to use the right amount of decimals this time 😁

shelbyd:
It's easy to run a local node with this spec. So we can check to make sure we have the right units, but I'm pretty confident we do.

PR #55, opened 2021-10-02
Health check sidecar cmd for substrate nodes

boymaas:


melatron (approved):

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        let healthy = healthy.clone();
+        warp::path!("health")
+            .map(move || healthy.clone())
+            .and_then(return_health_status)
+    };
+
+    tokio::spawn(poller(opts.clone(), healthy));
+
+    warp::serve(health_check)
+        .run(([0, 0, 0, 0], opts.port))
```
shelbyd, line 105: If you start warp before the poller, you can avoid cloning the opts. Alternatively, pass a reference to the opts.

shelbyd: Obviously performance doesn't matter, but it's a smell to have a pile of clones and copies unnecessarily throughout the code.

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+            reqwest::get(&opts.node_rpc_endpoint)
+                .await?
+                .json::<Health>()
+                .await
+        };
+
+        match resp.await {
+            Ok(health) => {
+                let node_is_healthy = !health.is_syncing
+                    && (!health.should_have_peers || health.peers >= opts.min_peers);
```
shelbyd, line 64: Move this into a method on Health. Then you can write it more readably:

```rust
fn is_healthy(&self, required_peers: usize) -> bool {
  if self.is_syncing {
    return false;
  }

  if self.should_have_peers {
    return self.peers >= required_peers;
  }

  return true;
}
```

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::sync::atomic::Ordering;
+use std::{sync::Arc, time::Duration};
+
+use anyhow::Result;
+use reqwest::StatusCode;
+use serde::{Deserialize, Serialize};
+use structopt::StructOpt;
+use warp::Filter;
+
+/// Health struct returned by the RPC
```
shelbyd, line 12: Since these are doc comments, I think they're meant to end with punctuation. I could be wrong though.

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+    // The port to listen on
+    #[structopt(short, long, default_value = "9955")]
+    port: u16,
+}
+
+async fn poller(opts: Opt, healthy: Arc<AtomicBool>) {
+    loop {
+        let resp = async {
+            reqwest::get(&opts.node_rpc_endpoint)
+                .await?
```
shelbyd, line 56: I'm unclear as to the behavior of `?` here. I'd expect it to return from the function, but it seems to be returning from the block?

If it returns from the function, then your poller would stop if the request fails and would only hit the `Err` case later if specifically loading the JSON failed. So I think that if the node is unreachable, the poller would stop and everything would look fine on AWS (despite it clearly not being fine).

shelbyd: Apparently, according to this: https://rust-lang.github.io/async-book/07_workarounds/02_err_in_async_blocks.html. the `?` operator returns from the async block.

Probably good to explicitly mark the return type of this function as a never type, `!` when the `never_type` feature is stabilized. Or an enum with no variants until then. `enum Never {}`. That makes it clear that it's impossible for the function to return.

shelbyd (changes requested):

support/node-health-check/src/main.rs
boymaas: Yes this is a trick I started using, and see being used in the wild as an alternative to a try block, which is also an unstable feature in rust. It's just to keep the code a bit cleaner/readable, so I can use the double ? chain, and not have to do nested matches or nested ifs.

support/node-health-check/src/main.rs
boymaas: Yes, agree, this expresses the logic more readable. Instead of the boolean logic.

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+        Ok(StatusCode::OK)
+    } else {
+        Ok(StatusCode::NOT_FOUND)
+    }
+}
+
+#[tokio::main]
+async fn main() {
+    let opts = Opt::from_args();
+    let healthy = Arc::new(AtomicBool::new(true));
```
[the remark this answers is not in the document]
boymaas: Its the most simple solution in my view. The `Arc<Mutex<bool>>` incurs another await, and locking etc. The AtomicBool does not require this, and since its Ordering is not important we can just use the Relaxing.

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    let health_check = {
+        let healthy = healthy.clone();
+        warp::path!("health")
+            .map(move || healthy.clone())
```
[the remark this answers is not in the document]
boymaas: the thing is, warp executes the move closure on each request, and delivers is to the `and_then` function as a parameter. This is why there has to be clone there, as is will create a new ref on each request.

support/node-health-check/src/main.rs
boymaas: Good point, I did this the first way around. I also do not like too many `clones`. It just does not feel good. In this case we do the await on `warp::serve` and the program blocks. We can do it, but then we have to `tokio::spawm` the `warp::serve` as well. This would avoid the clone. In this case I decided to go for the extra clone. Considering its not a time critical loop. But we think the same way on this point.

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::{sync::Arc, time::Duration};
+
+use anyhow::Result;
+use reqwest::StatusCode;
+use serde::{Deserialize, Serialize};
+use structopt::StructOpt;
+use warp::Filter;
+
+/// Health struct returned by the RPC
+#[derive(Debug, PartialEq, Serialize, Deserialize, Default)]
```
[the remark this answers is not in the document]
boymaas: Agree, will remove them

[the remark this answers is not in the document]
boymaas: Also removed the PartialEq

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+struct Opt {
+    /// Debug
+    #[structopt(short, long)]
+    debug: bool,
+
+    /// Set poll_interval
+    #[structopt(short = "i", default_value = "5")]
+    poll_interval: u64,
+
+    /// Set minimum peers
```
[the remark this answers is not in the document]
boymaas: I've added a comment describing what it does. It's in the logic of the is_healthy function. The node has to be connected to at least that many other peers if not in dev of local mode.

support/node-health-check/src/main.rs
boymaas: I fixed in an easier way, just doing the blocking await on the poller.

support/node-health-check/src/main.rs
boymaas: I searched a bit, but could not find what you mean with the punctuation. I will leave this one open.

support/node-health-check/src/main.rs
```diff
@@ -0,0 +1,113 @@
...
+struct Opt {
+    /// Debug
+    #[structopt(short, long)]
+    debug: bool,
+
+    /// Set poll_interval
+    #[structopt(short = "i", default_value = "5")]
+    poll_interval_in_secs: u64,
+
+    /// Set minimum peers to be connected for this node for it to be considered healthy
```
shelbyd, line 51: Still a little confusing, maybe "The inspected node must have at least this many peers to be considered healthy".

support/node-health-check/src/main.rs
shelbyd: Can you make the return type of the `poller` fn `Never` to make this more explicit?

support/node-health-check/src/main.rs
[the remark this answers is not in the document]
shelbyd: You can still do a blocking lock in an async function. It would be a problem if this were going to be highly async with many in progress requests. I'd still do `Arc<Mutex<bool>>` but I'm fine with this.

support/node-health-check/src/main.rs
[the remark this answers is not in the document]
shelbyd: If you have return health status take a `&Arc`, or possibly even `&AtomicBool` then it can avoid the clone. I'd think this would work:
```rust
{
  let healthy = healthy.clone();
  warp::path!("health").and_then(|_| return_health_status(&healthy))
}
```

shelbyd (approved):

support/node-health-check/src/main.rs
boymaas: Agree, more succint, changed it

support/node-health-check/src/main.rs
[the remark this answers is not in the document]
boymaas: It would be great if that would work, but it does not. I get this message: 
image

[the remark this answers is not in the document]
boymaas: and when I use the move it again complaints about the lifetimes:

image

support/node-health-check/src/main.rs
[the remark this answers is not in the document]
boymaas: Yes I know this, I have used the RwLock and Mutexes extensively. It's just a matter of taste, this will never deadlock, but in this pattern the `lock().await.write(..)` probably will also never deadlock.

support/node-health-check/src/main.rs
boymaas: Yes, I like it! I already stripped the Result from the `poller` to communicate no `?` could be used outside the block I was using it. This makes it even more explicit for future implementors.

PR #56, opened 2021-10-08
Add @fractal/polkadot-utils library with TxnBatcher class capable of ~20 txn/s

shelbyd:
Also includes a TxnWatcher class that encapsulates the logic of watching for different transaction statuses.

### DIFF
--- a/support/@fractal/polkadot-utils/package.json
+++ b/support/@fractal/polkadot-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fractal/polkadot-utils",
-  "version": "1.0.0",
+  "version": "0.1.0",
   "description": "Utilities to make interacting with Substrate nodes robust and efficient.",
   "main": "build/main/index.js",
   "typings": "build/main/index.d.ts",


support/@fractal/polkadot-utils/package.json
```diff
@@ -0,0 +1,104 @@
...
+    "timeout": "60s",
+    "typescript": {
+      "rewritePaths": {
+        "src/": "build/main/"
+      }
+    },
+    "files": [
+      "!build/module/**"
+    ]
+  },
```
melatron, line 89: Are we planning to use ava in the future, or this just came with the starting template?

Cargo.toml
```diff
@@ -3,5 +3,5 @@ members = [
   'blockchain/node',
   'blockchain/pallets/*',
   'blockchain/runtime',
-  'support/*',
+  'support/merklex',
```
melatron, line 6: Maybe it would be better to add `exclude = ["support/@fractal"]`

support/@fractal/polkadot-utils/package.json
shelbyd: Just came with the template. I'm happy to change it to anything else.

support/@fractal/polkadot-utils/examples/high_throughput_tranfers.js
```diff
@@ -0,0 +1,66 @@
+const {ApiPromise, Keyring, WsProvider} = require('@polkadot/api');
+const {TxnBatcher} = require('../build/main');
+
+async function isolatedBatcher() {
+  const provider = new WsProvider('ws://localhost:9944');
+  const api = await ApiPromise.create({provider});
```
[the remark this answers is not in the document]
shelbyd: We only need to if we ever try to use those types. Since this is a generic library, it should be usable against any Substrate runtime. We still get a warning when running locally, but it doesn't prevent anything from working.

PR #59, opened 2021-10-12
add node production aws sns cron

melatron:


support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+async function main() {
+    const api = await createPromiseApi(settings.nodeAddress);
+
+    let lastHeaderHex = settings.lastHeaderHex;
+    let counter = 0;
+    let sendingSns = false;
+    const job = nodeCron.schedule(settings.cronTime, () => {
+        console.log(new Date().toLocaleString());
```
shelbyd, line 33: Should log what we're doing. `console.log(new Date().toISOString(), 'Checking for new block');`. Or use a proper logging library that will format with relevant context.

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,62 @@
...
+            if (currentHeader.toHex() != lastHeaderHex) {
+                lastHeaderHex = currentHeader.toHex();
+                counter = 0;
+            } else {
+                counter++;
+                if (counter > 1 && !sendingSns) {
+                    sendingSns = true;
+
+                    // Create promise and SNS service object
+                    publishToTopic(settings.topicName, `message: ${settings.message};  Last header hex: ${lastHeaderHex}`).then((data) => {
```
shelbyd, line 45: Why not `await publishToTopic(...)`?

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,62 @@
...
+    return sns.publish({
+        Message: message,
+        TopicArn: topicArn,
+    }).promise();
+}
+
+async function main() {
+    const api = await createPromiseApi(settings.nodeAddress);
+
+    let lastHeaderHex = settings.lastHeaderHex;
```
shelbyd, line 29: Doesn't seem necessary to me to save and load these headers. I'd just start this script with getting the latest header.

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,62 @@
...
+                counter++;
+                if (counter > 1 && !sendingSns) {
+                    sendingSns = true;
+
+                    // Create promise and SNS service object
+                    publishToTopic(settings.topicName, `message: ${settings.message};  Last header hex: ${lastHeaderHex}`).then((data) => {
+                        sendingSns = false;
+                        counter = 0;
+                        console.log(`Message ${settings.message} sent to the topic ${settings.topicName}`);
+                        console.log("MessageID is " + data.MessageId);
```
shelbyd, line 49: Maybe want `console.warn` if we're notifying about alerts?

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,62 @@
...
+    const provider = new WsProvider(nodeAddress);
+    const api = await new ApiPromise({ provider, types });
+    await api.isReady;
+    return api;
+}
+
+async function publishToTopic(topicName, message) {
+    AWS.config.update({region: settings.region});
+    const sns = new AWS.SNS({apiVersion: settings.awsApiVersion});
+    const topics = await sns.listTopics().promise();
```
shelbyd, line 17: How long does this RPC take? If it's long, we may want to do this when the script starts to avoid additional latency when detecting problems.

shelbyd (changes requested):

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,55 @@
...
+            console.log(new Date().toISOString(), 'Checking for new block');
+            const signedBlock = await api.rpc.chain.getBlock();
+            const currentHeader = signedBlock.block.header.hash;
+
+            if (currentHeader.toHex() != lastHeaderHex) {
+                lastHeaderHex = currentHeader.toHex();
+                lastNewBlockAt = Date.now();
+                return;
+            }
+            console.log(`${Date.now()} , ${lastNewBlockAt}, ${Settings.requireNewBlockEveryMs}`)
```
shelbyd, line 36: Please format messages so we know what is going on just reading the logs. I think this would be something like:

```
console.log(new Date().toISOString(), `Have not seen new block for ${new Date() - lastNewBlockAt}ms`);
```

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,55 @@
...
+            if (!sendingSns && (Date.now() - lastNewBlockAt > Settings.requireNewBlockEveryMs)) {
+                sendingSns = true;
+                let data = await sns.publish({
+                    Message: `message: ${Settings.message};  Last header hex: ${lastHeaderHex}`,
+                    TopicArn: topicArn,
+                }).promise();
+                console.warn(`Message ${Settings.message} sent to the topic ${topicArn}`);
+                console.warn("MessageID is " + data.MessageId);
+                sendingSns = false;
+            }
```
shelbyd, line 46: If I understand correctly, this would repeatedly send notifications as long as blocks aren't being produced. After successfully sending a notification, we should wait until we see a new block to start checking again (or around 30 minutes).

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,55 @@
...
+            const signedBlock = await api.rpc.chain.getBlock();
+            const currentHeader = signedBlock.block.header.hash;
+
+            if (currentHeader.toHex() != lastHeaderHex) {
+                lastHeaderHex = currentHeader.toHex();
+                lastNewBlockAt = Date.now();
+                return;
+            }
+            console.log(`${Date.now()} , ${lastNewBlockAt}, ${Settings.requireNewBlockEveryMs}`)
+            if (!sendingSns && (Date.now() - lastNewBlockAt > Settings.requireNewBlockEveryMs)) {
```
shelbyd, line 37: Maybe `if (sendingSns) return`?

support/node-production-aws-sns/index.js
melatron: I used this `console.log` for debug purpose, but I'll leave it with some better formatting.

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,65 @@
...
+                lastNewBlockAt = Date.now();
+                console.log(`Block header hex: ${lastHeaderHex} at ${new Date().toISOString()}`);
+                continue;
+            }
+
+            console.log(new Date().toISOString(), `Have not seen new block for ${new Date() - lastNewBlockAt}ms`);
+
+            if (sendingSns) {
+                continue;
+            }
```
shelbyd, line 43: With this being in an async loop, I don't think it's possible for this to be true now. It was possible with node-cron, but now I don't think we need to track this.

support/node-production-aws-sns/index.js
```diff
@@ -0,0 +1,65 @@
...
+
+            if (Date.now() - lastNewBlockAt > Settings.requireNewBlockEveryMs) {
+                sleepTime = 1000 * 60 * Settings.minutesUntilNextCheckAfterAlarm;
```
shelbyd, line 46: How about moving this to the end of the block and sleeping the amount directly. Since this is the only place we change sleepTime, then we wouldn't need to have a local variable and can have the initial sleep always do what's configured.

PR #61, opened 2021-10-13
Switch to new mainnet spec addresses

shelbyd:


juliosantos (approved):
please update the sheet with the new keys

PR #63, opened 2021-10-13
Bmaas/mainnet reset

boymaas:
Script to run a mainnet reset


scripts/mainnet/reset/rebuild-spec.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -x
+
+NODE_BUILD=$1
```
shelbyd, line 3: This argument should be the whole tag (currently would be `bmaas/nodefcl:<something>`). One less thing to change when we move where we're pushing images to.

scripts/mainnet/reset.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash -x
+
+usage() {
+  cat << EOT
+Usage: $0 <<BUILD ID>>
+Reset fractals mainnet, example BUILD ID = e1ca4db_20211011_1204
```
shelbyd, line 6: "Reset Fractal Mainnet."

"Example:
  $0 trustfractal/protocol-node:1234"

scripts/mainnet/reset.sh
```diff
@@ -0,0 +1,75 @@
...
+# /p2p/12D3KooWB3ENNs5vK5HNJTHVd33wEZnXdoTFwHkEUcFdQ6JxRhRJ
+subsitute_bootnode_id() {
+  ssh $1 "PERL_BADLANG=0 perl -pi -e \"s|/p2p/\w+|/p2p/$2|\" docker-compose.yml"
+}
+
+(( $# < 1 )) && { echo "1 arguements is required."; usage; exit 1; }
+
+NODE_BUILD_ID=$1
+
+node_00="ubuntu@node-boot.mainnet.fractalprotocol.com"
```
shelbyd, line 23: Since we treat this node differently, we should probably have the name different from the others.

scripts/mainnet/reset.sh
```diff
@@ -0,0 +1,75 @@
...
+# change the docker-compose files splice in the new node-build-id for $node_01 and $node_02
+# run the insert-keys step for all the nodes
+subsitute_bootnode_id $node_01 $bootnode_id
+subsitute_bootnode_id $node_02 $bootnode_id
+
+ssh $node_01 'docker-compose up -d'
+ssh $node_02 'docker-compose up -d'
+
+# now shoot in keys and reboot
+echo "Now shoot in the prod and val keys ..  "
```
shelbyd, line 70: This isn't super clear what the script is asking of the user. Probably something like "Upload block authoring and finality keys <link to substrate page>".

scripts/mainnet/reset.sh
```diff
@@ -0,0 +1,75 @@
...
+
+usage() {
+  cat << EOT
+Usage: $0 <<BUILD ID>>
+Reset fractals mainnet, example BUILD ID = e1ca4db_20211011_1204
+EOT
+}
+
+subsitute_node_build_id() {
+  ssh $1 "PERL_BADLANG=0 perl -pi -e \"s/\\\\w+_\\\\d+_\\\\d+/$2/\" docker-compose.yml"
```
shelbyd, line 11: Would it be simpler to use ENV variable substitution in the docker-compose file? https://docs.docker.com/compose/environment-variables/

scripts/mainnet/reset.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash -x
+
```
shelbyd, line 2: Add at least `set -e` to have the script exit if one step fails.

I also usually add `u` and `o pipefail`. I don't remember what the u does though.

So `set -euo pipefail`.

scripts/mainnet/reset.sh
boymaas: Agree

scripts/mainnet/reset/rebuild-spec.sh
boymaas: I agree, but then we get a problem with the `substitute_` functions, there is some coupling here between what the value is in the `docker-compose.yml` to be replaced (at the bmaas/nodefcl tag is also present) there. Maybe leave this in and just do a search replace over multiple files. Will revisit this tomorrow.

PR #64, opened 2021-10-13
Data host boilerplate

melatron:


.github/workflows/data-host.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/cache@v2
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - name: Install packages
+        run: yarn install
+      - name: Run tests
+        run: yarn test
+      #  run: yarn run cov:send
+      #  run: yarn run cov:check
```
shelbyd, line 40: FWIW, I have no interest in including code coverage. At best, it's a slightly interesting metric, and at worst it makes teams test things they shouldn't and not cover all interesting cases.

support/@trustfractal/data-host/.vscode/launch.json
```diff
@@ -0,0 +1,29 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    // To debug, make sure a *.spec.ts file is active in the editor, then run a configuration
```
shelbyd, line 4: Unfortunately, most JSON parsers don't allow comments. So we should remove these.

support/@trustfractal/data-host/.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
+    "eamodio.gitlens",
+    "streetsidesoftware.code-spell-checker",
+  ]
+}
```
shelbyd, line 8: I personally don't use vscode. It also seems strange to include editor specific config in source control. I'd prefer not to have this unless there's a good reason to.

support/@trustfractal/data-host/README.md
```diff
@@ -0,0 +1,3 @@
+# data-host
+
+Javascript library for data providers using fractal protocol
```
shelbyd, line 3:
```suggestion
# @trustfractal/data-host

JavaScript library for Chrome Extensions looking to include a Fractal Protocol Data Host in their extension.
```

shelbyd (approved):

support/@trustfractal/data-host/.vscode/extensions.json
melatron: I am using vscode and thought it would be useful, but I can add it as a local configuration.

PR #65, opened 2021-10-13
Add fractal properties for the chainspec

melatron:


blockchain/node/src/chain_spec.rs
```diff
@@ -39,9 +39,17 @@ pub fn authority_keys_from_seed(s: &str) -> (AuraId, GrandpaId) {
     (get_from_seed::<AuraId>(s), get_from_seed::<GrandpaId>(s))
 }
 
+pub fn fractal_properties() -> Properties {
+    let mut p = Properties::new();
+    // p.insert("ss58format".into(), <number>.into());
```
shelbyd, line 44: Don't include commented code.

```suggestion
```

blockchain/node/src/chain_spec.rs
```diff
@@ -76,7 +84,7 @@ pub fn development_config() -> Result<ChainSpec, String> {
         // Protocol ID
         None,
         // Properties
-        None,
+        Some(props),
```
shelbyd, line 87: We shouldn't need a variable for this, just `Some(fractal_properties())` is good. Same in mainnet_config.

shelbyd (changes requested):

PR #67, opened 2021-10-15
Add runtime upgrade script

melatron:


Cargo.toml
```diff
@@ -8,5 +8,6 @@ members = [
 
 exclude = [
   'support/@trustfractal',
+  'support/runtime-upgrade',
```
shelbyd, line 11: This should be in `/scripts`.

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+    const api = await createPromiseApi(nodeAddress);
+    const keyring = new Keyring({ type: 'sr25519' });
+    // Retrieve the upgrade key from the chain state
+    const adminId = await api.query.sudo.key();
+
+    // Some mnemonic phrase
+    const PHRASE = fs.readFileSync(argv.rootMnemonicPath).toString().trimEnd();
+
+    // Add an account, straight mnemonic
+    const newPair = keyring.addFromUri(PHRASE);
```
shelbyd, line 25: Should probably check and error if the provided key doesn't match the Sudo key.

support/runtime-upgrade/README.md
```diff
@@ -0,0 +1,3 @@
+# Simple JS tool for substrate upgrades
+
+Eaxmple: `./support/runtime-upgrade/upgrade_chain.sh ws://127.0.0.1:9944 ./support/runtime-upgrade/example_mnemonic`
```
shelbyd, line 3: Don't include the directory the script is in for examples like this. It adds unnecessary noise and it would be incorrect if the file is moved. Consumers can figure out where to run it from.

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
+const { ApiPromise, WsProvider, Keyring } = require('@polkadot/api');
+
+var argv = require('minimist')(process.argv.slice(2));
```
shelbyd, line 3: Your example just shows providing arguments in sequence, doesn't minimist require arguments to be named, like `--node-address wss://...`?

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+        : api.tx.consensus.setCode(`0x${code}`); // For previous versions
+
+    console.log(`Upgrading from ${adminId}, ${code.length / 2} bytes`);
+
+    // Perform the actual chain upgrade via the sudo module
+    api.tx.sudo.sudoUncheckedWeight(proposal, 1).signAndSend(newPair, ({ events = [], status }) => {
+        console.log('Proposal status:', status.type);
+
+        if (status.isInBlock) {
+            console.error('You have just upgraded your chain');
```
shelbyd, line 40: Why error?

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+            console.log('Included at block hash', status.asInBlock.toHex());
+            console.log('Events:');
+
+            console.log(JSON.stringify(events, null, 2));
+        } else if (status.isFinalized) {
+            console.log('Finalized block hash', status.asFinalized.toHex());
+
+            process.exit(0);
+        }
```
shelbyd, line 50: If there's an error, this will just hang forever. Consider using the TxnWatcher class from `@trustfractal/polkadot-utils`. You may need to make a static `TxnWatcher.signAndSend(txn, signer, api)` method.

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+}
+
+main().catch((error) => {
+    console.error(error);
+    process.exit(-1);
```
shelbyd, line 56: I thought exit codes needed to be positive? This should be `exit(1)` I think.

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+    // Perform the actual chain upgrade via the sudo module
+    api.tx.sudo.sudoUncheckedWeight(proposal, 1).signAndSend(newPair, ({ events = [], status }) => {
+        console.log('Proposal status:', status.type);
+
+        if (status.isInBlock) {
+            console.error('You have just upgraded your chain');
+
+            console.log('Included at block hash', status.asInBlock.toHex());
+            console.log('Events:');
```
shelbyd, line 43: ?

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+    await api.isReady;
+    return api;
+}
+
+async function main() {
+    const nodeAddress = argv.nodeAddress;
+    const api = await createPromiseApi(nodeAddress);
+    const keyring = new Keyring({ type: 'sr25519' });
+    // Retrieve the upgrade key from the chain state
+    const adminId = await api.query.sudo.key();
```
shelbyd, line 19: This is declared pretty far away from where it is used (logically speaking). I'd recommend declaring things as late a possible for a few reasons:

1. Limits what readers need to think about.
2. Communicates the scope of the relevance of the variable better.
3. Keeps failures/messages more in context. For example, if we couldn't complete the RPC and you specified a bad path for the mnemonic, we would fail first with the RPC, when we didn't actually need to yet.

support/runtime-upgrade/example_mnemonic
```diff
@@ -0,0 +1 @@
+//Alice
```
shelbyd, line 1: This isn't a mnemonic. The script takes any Substrate private key (mnemonic, derivation path, etc). The argument and docs including 'mnemonic' could make users think they needed to provide a mnemonic, when any key would work.

I'd change any reference to 'mnemonic' to privateKey or something. Also this file should have an extension, something like `path/key.priv`.

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
+const { ApiPromise, WsProvider, Keyring } = require('@polkadot/api');
+
+var argv = require('minimist')(process.argv.slice(2));
+const fs = require('fs');
+
+// Create a promise API instance of the passed in node address.
+async function createPromiseApi(nodeAddress) {
+    const wsProvider = new WsProvider(nodeAddress);
+    const api = await new ApiPromise({ provider: wsProvider });
+    await api.isReady;
```
shelbyd, line 10: I think all API methods await this for us. I think you can just `await new ApiPromise`.

support/runtime-upgrade/index.js
```diff
@@ -0,0 +1,57 @@
...
+    const PHRASE = fs.readFileSync(argv.rootMnemonicPath).toString().trimEnd();
+
+    // Add an account, straight mnemonic
+    const newPair = keyring.addFromUri(PHRASE);
+
+    // Retrieve the runtime to upgrade
+    const code = fs.readFileSync(argv.wasmPath).toString('hex');
+    const proposal = api.tx.system && api.tx.system.setCode
+        ? api.tx.system.setCode(`0x${code}`) // For newer versions of Substrate
+        : api.tx.consensus.setCode(`0x${code}`); // For previous versions
```
shelbyd, line 31: We don't run previous versions, we should only include the code for the version we run.

support/runtime-upgrade/index.js
shelbyd: It would also make this code cleaner:

```js
const txn = TxnWatcher.signAndSend(api.tx.sudo.sudoUncheckedWeight(proposal, 1), newPair);
const { block } = await txn.inBlock();
console.log('Included in block', block.toHex());
const { block } = await txn.finalized();
console.log('Finalized at', block.hash);
```

support/runtime-upgrade/index.js
shelbyd: Should be `newPair.address`.

support/runtime-upgrade/index.js
shelbyd: That's very strange, I haven't had to do that anywhere else.

scripts/runtime-upgrade/upgrade_chain.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
+SCRIPT_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+(cd $SCRIPT_DIR../../ ; cargo build --release)
```
shelbyd, line 3: Consider making a `ROOT_DIR` variable since you use `$SCRIPT_DIR/../../` in multiple places. Less likely to miss updating one if moving this file.

scripts/runtime-upgrade/index.js
```diff
@@ -0,0 +1,47 @@
...
+    console.log(`Upgrading from ${adminId}, ${code.length / 2} bytes`);
+
+    // Perform the actual chain upgrade via the sudo module
+    const txn = TxnWatcher.signAndSend(api.tx.sudo.sudoUncheckedWeight(proposal, 1), newPair);
+    await txn.inBlock();
+    console.log('Included in block'); //TODO: include the block hash when TxnWatcher is updated to return it from `inBlock`
+    //TODO: Also we can log the events here
+    await txn.finalized();
+    console.log('Finalized'); //TODO: include the finalized block hash when TxnWatcher is updated to return it from `inBlock`
+    process.exit(1);
```
shelbyd, line 41: Should exit 0 on success. And I think just finishing the main function will cause node to exit(0).

shelbyd (approved):

PR #70, opened 2021-10-16
Add utility to generate exponential to linear issuance curve

shelbyd:
Sample issuance curve generated by this code: https://docs.google.com/spreadsheets/d/1cz-F437zp9vaVoZ44rLj5UFSW3xptSm414QxEe8kfYc/edit#gid=1281634863

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1639,6 +1639,7 @@ dependencies = [
  "frame-benchmarking",
  "frame-support",
  "frame-system",
+ "libm",
  "merklex",
  "num-traits",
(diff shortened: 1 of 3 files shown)


juliosantos:
> generated by this code

I don't understand the period in column A. If days, it takes ~30 years. If weeks, ~210.

We're supposed to mint 2,302,047 FCL in the first month, and the remaining 397,697,953 over 50+ years. Rolling up the first 7 periods yields 2,125,912, and the first 8 yield 2,733,142.

PR #76, opened 2021-10-19
Add types and mainnet address

melatron:


support/node-production-aws-sns/index.js
```diff
@@ -1,11 +1,12 @@
 const { ApiPromise, WsProvider } = require('@polkadot/api');
 const AWS = require('aws-sdk');
 const Settings = require('./settings.json')
+const types = require('../../blockchain/types.json')
```
shelbyd, line 4: I'm not a fan of relative imports, but configuring for absolute imports seems difficult... Consider researching some ways to make this import independent of where the node-production-aws-sns file is located.

support/node-production-aws-sns/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "nodeAddress": "ws://127.0.0.1:9944",
+    "nodeAddress": "wss://nodes.mainnet.fractalprotocol.com",
```
shelbyd, line 2: Nit: Seems like something that should bee a command line argument to the script (perhaps defaulting to the mainnet).

shelbyd (approved):

support/node-production-aws-sns/index.js
melatron: Maybe I can request the json file from the repo? But this doesn't seem right... 
`request('https://github.com/trustfractal/protocol/blob/master/blockchain/types.json', (error, response, body) => { ... })`

shelbyd: Definitely not. Relative import is much better than a network request and encoding the repository.

PR #78, opened 2021-10-25
Return block hex to inBlock and finalized

melatron:


support/@trustfractal/polkadot-utils/examples/high_throughput_tranfers.js
```diff
@@ -49,15 +49,15 @@ async function main() {
...
       })());
     }, 10)
   })
 
-  await Promise.all(promises);
-  console.log('All txns in block');
-  await Promise.all(Array.from(watchers).map(w => w.finalized()));
-  console.log('All txns finalized');
+  let inBlocks = await Promise.all(promises);
+  console.log('All txns in block', inBlocks);
```
shelbyd, line 58: I think this will be a really long list of repeated block hashes. Probably better as a map between block-hash and number in block? Similar to how we do statuses above.

support/@trustfractal/polkadot-utils/src/lib/watcher.ts
```diff
@@ -12,8 +12,8 @@ export class TxnWatcher {
   public status: AnyJson | string = 'Unsubmitted';
 
   private onReady = new MultiCallback<void>();
-  private onInBlock = new OnceMultiCallback<void>('onInBlock');
-  private onFinalized = new OnceMultiCallback<void>('onFinalized');
+  private onInBlock = new OnceMultiCallback<string>('onInBlock');
```
shelbyd, line 15: These should be an object so introducing additional data isn't a breaking change. Something like:

```ts
export interface TxnInBlock {
  block: string,
  txnHash: string,
}
```

But we only need the block hash right now. Same for OnFinalized.

support/@trustfractal/polkadot-utils/src/lib/watcher.ts
```diff
@@ -38,12 +38,12 @@ export class TxnWatcher {
         // unhandled case below.
       } else if (result.status.isInBlock) {
         this.status = 'InBlock';
-        this.onInBlock.callAll();
+        this.onInBlock.callAll(result.status.asInBlock.toHex());
       } else if (result.status.isFinalized) {
-        this.onInBlock.callIfUncalled();
+        this.onInBlock.callIfUncalled(result.status.asFinalized.toHex());
```
shelbyd, line 43: Is `asFinalized` the same hash as `asInBlock`? Otherwise, we could call this with the incorrect InBlock hash.

shelbyd (changes requested):

support/@trustfractal/polkadot-utils/src/lib/watcher.ts
melatron: I am not sure, but when the status is `isFinalized` the `result.status.asInBlock` doesn't exists. That is why I assumed it should be the correct hash.

shelbyd: If you log them both with the example script that should let you know if they're the same.

support/@trustfractal/polkadot-utils/src/lib/watcher.ts
```diff
@@ -38,12 +46,12 @@ export class TxnWatcher {
         // unhandled case below.
       } else if (result.status.isInBlock) {
         this.status = 'InBlock';
-        this.onInBlock.callAll();
+        this.onInBlock.callAll( { block: result.status.asInBlock.toHex() } );
```
shelbyd, line 49: This formatting looks different from other files, did you run a formatter on this?

shelbyd (approved):

PR #86, opened 2021-11-09
Initial commit for data-host InjectedScript and Background setup.

melatron:

PR #88, opened 2021-11-16
[WIP] Data host opt in

melatron:


shelbyd:
If this depends on #86, please make the base of this PR that branch. It makes it much easier for me to review since I see only what is new in this PR.

data_host/chrome_extension/src/pages/Popup/index.tsx
```diff
@@ -1,23 +1,18 @@
-import React from 'react';
-import { render } from 'react-dom';
-import { Provider } from 'react-redux';
+// import React from 'react';
+// import { render } from 'react-dom';
+// import { Provider } from 'react-redux';
```
shelbyd, line 3: Please don't comment code (unless it's for documentation).

support/@trustfractal/data-host/src/popup/Environment.ts
```diff
@@ -0,0 +1,25 @@
...
+  'development';
+
+const environment: Environment = {
+  FRACTAL_WEBSITE_URL,
+  IS_DEV: NODE_ENV === 'development',
+  PROTOCOL_RPC_ENDPOINT,
+  PROTOCOL_CURRENCY,
+};
+
+export default environment;
```
shelbyd, line 25: We should expose these to clients as configuration with production values being the default. Clients will have their own means of configuring environments, and we shouldn't impose.

support/@trustfractal/data-host/src/popup/Environment.ts
```diff
@@ -0,0 +1,25 @@
...
+const FRACTAL_WEBSITE_URL =
+  //   process.env.REACT_APP_FRACTAL_WEBSITE_URL ||
+  'https://staging.sandbox.fractal.id';
+const PROTOCOL_RPC_ENDPOINT =
+  //   process.env.REACT_APP_PROTOCOL_RPC_ENDPOINT ||
+  'wss://nodes.testnet.fractalprotocol.com';
+
+const PROTOCOL_CURRENCY =
+  // process.env.REACT_APP_PROTOCOL_CURRENCY ||
+  'FCL';
```
shelbyd, line 12: Good to remove.

support/@trustfractal/data-host/src/popup/Environment.ts
```diff
@@ -0,0 +1,25 @@
+import { Environment } from '@popup/types/Environment';
+
+const FRACTAL_WEBSITE_URL =
+  //   process.env.REACT_APP_FRACTAL_WEBSITE_URL ||
```
shelbyd, line 4: Don't comment out code.

support/@trustfractal/data-host/src/popup/components/App.tsx
```diff
@@ -1,9 +1,12 @@
...
+import Protocol from '@components/Protocol';
+import GlobalStyle from "@styles/GlobalStyle";
 function App() {
 
   return (
+    <>
+    <GlobalStyle />
     <Protocol />
+    </>
   );
```
shelbyd, line 10: Do we have a formatter configured that handles jsx? I think these are usually indented below the top-level <></>.

support/@trustfractal/data-host/src/popup/components/Protocol.tsx
```diff
@@ -1,10 +1,63 @@
...
+
+        setPageOverride(
+          <SetupSuccess
+            mnemonic={mnemonic}
+            onContinue={() => setPageOverride(null)}
+          />,
+        );
+      } catch (e) {
+        console.log(e);
+        // handleError(e, () => optInWithMnemonic(mnemonic));
```
shelbyd, line 41: Why remove this?

support/@trustfractal/data-host/src/popup/services/DataHost.ts
```diff
@@ -0,0 +1,96 @@
+import { Storage, StorageArray } from '@utils/StorageArray';
+
+export class DataHost {
```
shelbyd, line 3: This is the type of thing we should write in Rust. If we want to include it as TS for now, just to make progress, that's fine. But we should move this into a Rust crate.

support/@trustfractal/data-host/package.json
```diff
@@ -1,25 +1,24 @@
 {
   "name": "@trustfractal/data-host",
```
shelbyd, line 2: I'd like to have the Data Host library in /data_host/library or something similar. I think /support should really only be for very minimal focused things that we use to support our own development.

support/@trustfractal/data-host/src/popup/services/Factory.ts
```diff
@@ -0,0 +1,106 @@
...
+import { ProtocolService } from '@services/Protocol';
+import { ProtocolOptIn } from '@services/ProtocolOptIn';
+import { StorageService } from '@services/StorageService';
+import { AggregateMultiContext, MultiContext } from '@utils/MultiContext';
+import { ValueCache } from '@utils/ReactHooks';
+
+const types = {
+  FractalId: 'u64',
+  MerkleTree: 'Raw',
+};
```
shelbyd, line 14: Maybe later: We should import these from our source of truth, /blockchain/types.json I think.

support/@trustfractal/data-host/src/popup/services/Factory.ts
```diff
@@ -0,0 +1,106 @@
...
+
+let storageService: StorageService;
```
shelbyd, line 16: I'm not super happy with having this one factory file. I'd prefer a dependency injection framework, but I didn't want to do the effort of integrating that into the wallet. We can consider changing this.

support/@trustfractal/data-host/src/popup/services/FractalAccount.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(message?: string) {
+    super(message);
+    this.name = 'NotConnectedError';
+  }
+}
+
+const NEXT_TOKENS_KEY = 'fractal-account-connector/will-accept-next-tokens';
+const TOKENS_KEY = 'fractal-account-connector/tokens';
+
+export class FractalAccountConnector extends MultiContext {
```
shelbyd, line 20: Most of this class is for our old way of connecting the account. We can remove/change it now or later.

support/@trustfractal/data-host/src/popup/services/StorageService/Errors.ts
```diff
@@ -0,0 +1,79 @@
+enum ErrorCode {
```
shelbyd, line 1: We shouldn't need this file.

support/@trustfractal/data-host/src/popup/services/StorageService/index.ts
```diff
@@ -0,0 +1,74 @@
+import {
+  ERROR_CLEAR,
+  ERROR_GET_ITEM,
+  ERROR_HAS_ITEM,
+  ERROR_REMOVE_ITEM,
+  ERROR_SET_ITEM,
+} from './Errors';
```
shelbyd, line 7: Let's change how this service works to use subclasses of Error like the other services.

shelbyd (approved):

support/@trustfractal/data-host/src/popup/services/FractalAccount.ts
melatron: By the old way, do you mean that we are currently using another way in the Fractal Wallet? If so, can you send me link to it.

support/@trustfractal/data-host/src/popup/services/Factory.ts
melatron: Do you have on mind a specific dependency injection framework, or should I research?

shelbyd: I like Angular's, but I don't know how usable it is outside of Angular projects. Not worth holding up this PR though.

support/@trustfractal/data-host/src/popup/services/FractalAccount.ts
shelbyd: I mean the way the wallet is using now, but what we're moving away from. You can include this for now, but be sure to remove the unused code once fractal.id registers the identity.

PR #91, opened 2021-12-01
Ingest basic chain data into storage that Rust indexers will read

shelbyd:
There isn't a good library for loading Substrate chain data in Rust. So we'll do just that in JS.


explorer/src/ingest.js
```diff
@@ -0,0 +1,224 @@
...
+    await this.setKey(`block/${blockNumber}/extrinsic/${index}`, json);
+  }
+}
+
+async function updateLargestBlock(api, storage) {
+  while (true) {
+    const latestBlockNumber = (await api.rpc.chain.getBlock()).block.header.number.toNumber();
+    await storage.setLargestBlock(latestBlockNumber);
+    console.log(`Largest block ${latestBlockNumber}`)
+    await sleep(6000);
```
melatron, line 105: Maybe make the sleep duration configurable.

melatron (approved):

explorer/src/ingest.js
shelbyd: I made it not sleep at all, and have it do work as soon as there's a new block. Much better solution than sleeping at all.

PR #93, opened 2021-12-02
Capture sum of unique identities in identities indexer

shelbyd:


explorer/src/indexing/identities.rs
```diff
@@ -4,10 +4,19 @@ use crate::data::Extrinsic;
...
 pub struct CountIdentities {
     current_block: Option<u64>,
+    previous_count: i32,
+    this_count: i32,
 }
 
 impl Indexer for CountIdentities {
-    fn begin(&mut self, pg: &mut Client) -> anyhow::Result<()> {
+    fn version(&mut self) -> u32 {
+        2
```
melatron, line 13: Shouldn't this be in some kind of config file?

melatron (approved):

explorer/src/indexing/identities.rs
shelbyd: No. This is a value that will cause the top level manager to rerun the indexer. This is entirely driven by the code that produces the index. I'll add some docs to the trait to explain.

PR #95, opened 2021-12-07
[WIP] Add Data Screen

melatron:


shelbyd (approved):
Please run an unused function checker and remove any unused functions. Also fine to wait until we pull in most everything then do it all at once. But we should make sure to do it. Also might be good to add a lint to the automated build.

PR #96, opened 2021-12-09
Improve indexing performance by ~2x

shelbyd:
Also store block and extrinsic JSON in dedicated tables. Keeping them in a simple key/value store makes it difficult to extract relevant sections of keys since they are lexicographically ordered.

### DIFF
--- a/explorer/src/bin/index.rs
+++ b/explorer/src/bin/index.rs
@@ -17,6 +17,10 @@ struct Options {
 
     #[structopt(long, default_value = "2000")]
     no_blocks_wait_ms: u64,
+
+    // Default 200MB
+    #[structopt(long, default_value = "200000000")]
+    cache_byte_limit: usize,
 }
 
 impl Options {
@@ -33,6 +37,12 @@ fn main() -> anyhow::Result<()> {
 
     let indexers = indexing::indexers();
 
+    let lru_storage = shared_lru::SharedLru::with_byte_limit(options.cache_byte_limit);
+    let ingested = Arc::new(ingested::Ingested::create(
+        &lru_storage,
+        options.postgres()?,
+    ));
+
     let (error_tx, error_rx) = channel();
     let unpark = UnparkOnDrop::current();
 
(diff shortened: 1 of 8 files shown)


explorer/src/bin/index.rs
```diff
@@ -142,10 +154,10 @@ fn run_indexer(
     log::info!("Starting '{}'", id);
 
     indexer.begin(&mut pg)?;
-    let mut log_every = Interval::new(Duration::from_secs(1));
+    let mut one_second = Interval::new(Duration::from_secs(1));
```
melatron, line 157: When we include crate types like this:
```
use std::{
    sync::{atomic::*, mpsc::*, Arc},
    thread::*,
    time::*,
};
```
Isn't it better to explicitly provide all the ones that are going to be used? For example the `Interval`

explorer/src/ingested.rs
```diff
@@ -1,45 +1,134 @@
...
-) -> anyhow::Result<Option<crate::data::Extrinsic>> {
-    if let Some(value) = get_key(format!("block/{}/extrinsic/{}", block, index), pg)? {
-        Ok(Some(serde_json::from_str(&value)?))
-    } else {
-        Ok(None)
+impl Ingested {
+    pub fn create(lru: &Arc<SharedLru>, client: Client) -> Ingested {
+        Ingested {
+            blocks: lru.make_cache(),
+            extrinsics: lru.make_cache(),
```
melatron, line 29: Does this mean that both blocks and extrinsics would be cached? I guess that's why you've implemented `MemorySize` to them. My question is where exactly are you using the LRU cache? As I see we are getting specific `u64`s as key, or it would be the least recently used with this key?

melatron (approved):

explorer/src/bin/index.rs
shelbyd: It depends. For std/core, I don't mind glob including. But especially for unfamiliar 3rd party crates, I explicitly name all the imports I use. Unless there are a ton of imports, or only imports from a few crates. So it's a decision to make each time.

shelbyd: Also, Interval is defined in this file, not a std::time thing.

explorer/src/ingested.rs
shelbyd: Yes, both are cached. Calls to `load_block` and `load_extrinsic` pull from the cache first and only do a DB query if the cache is missing the value. The `shared_lru` crate will evict the least recently used item across all caches when full, so any block/extrinsic/something else could be evicted.

PR #102, opened 2022-01-02
Minting registration

melatron:


data-host/chrome-extension/src/pages/Background/index.js
```diff
@@ -1,12 +1,6 @@
...
-import { REDUX_PORT_NAME } from '../../redux/port';
-
-const store = createStore(reducers);
-wrapStore(store, { portName: REDUX_PORT_NAME });
+import { Background } from '@trustfractal/data-host';
 
 (() => {
-  console.log('Background ready');
+    const background = new Background();
+    background.addListeners();
```
shelbyd, line 5: Consider inlining: `new Background().addListeners()`.

shelbyd: Same elsewhere.

data-host/chrome-extension/src/pages/Popup/index.tsx
```diff
@@ -1,16 +1,9 @@
...
-const store = new Store({ portName: REDUX_PORT_NAME });
+let ui = new FractalUI();
+ui.render(document.getElementById('app-container'));
 
-store.ready().then(async () => {
-    let ui = new FractalUI();
-    ui.render(document.getElementById('app-container'));
-});
 
 if (module.hot) module.hot.accept();
```
shelbyd, line 9: We may need to reinitialize something for hot reloading to work? But NBD really.

data-host/chrome-extension/webpack.config.js
```diff
@@ -208,6 +208,15 @@ var options = {
...
   },
+//   experiments: {
+    // asyncWebAssembly: true,
+    // buildHttp: true,
+    // layers: true,
+    // lazyCompilation: true,
+    // outputModule: true,
+    // futureDefaults: true,
+    // topLevelAwait: true,
+//   },
```
shelbyd, line 219: Please don't leave commented code.

data-host/library/src/lib/background.ts
```diff
@@ -1,5 +1,36 @@
...
+  }
+  addListeners() {
+    //TODO(melatron): Create more generic method that handles all type of messages (for different facts)
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
+    chrome.runtime.onMessage.addListener(function (
+      request,
+      _sender,
+      sendResponse
+    ) {
```
shelbyd, line 31: Why is this a function and not a lambda? If you `addListener((request, ...) => {` you wouldn't need to `const self = this;`.

data-host/library/src/lib/injectionScript.ts
```diff
@@ -5,4 +5,10 @@ export class InjectionScript {
       `InjectionScript initialized with address: ${substrateAddress}.`
     );
   }
+  sendCurrentPageView() {
+    //TODO: send the fund to the background script
+    console.log('fractal fact: ', window.location);
```
shelbyd, line 10: remove?

data-host/library/src/popup/components/About.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <ContentContainer>
+        <Title>
+          Fractal Wallet{" "}
+          <Version>
+            <Text>{`v${version}`}</Text>
+          </Version>
+        </Title>
+        <Text>
+          <br />
+        </Text>
```
shelbyd, line 48: Is this necessary? If this is for spacing, please use actual CSS instead.

shelbyd: Same elsewhere.

data-host/library/src/popup/components/Alerts.tsx
```diff
@@ -1,9 +1,9 @@
-import styled from "styled-components";
-import { useState, useEffect } from "react";
 import { getUserAlerts } from "@services/Factory";
+import { useEffect, useState } from "react";
+import styled from "styled-components";
 
 export class UserAlerts {
-  public onMessage = (_: string) => {};
+  public onMessage = (_: string) => { console.log('onMessage')};
```
shelbyd, line 6: revert?

data-host/library/src/popup/components/DataScreen/WebpageViews.tsx
```diff
@@ -18,7 +18,7 @@ function HistoryList({ n }: { n: number }) {
   const history = useLoadedState(async () => {
     const items = [];
     for await (const item of getDataHost().iterBack()) {
-      if (item.pageView == null) continue;
+      if (!item || item.pageView == null) continue;
```
shelbyd, line 21: Did we change iterBack to sometimes return null? We shouldn't do that. The iter methods on data host should only return valid facts.

data-host/library/src/popup/components/SendTokens.tsx
```diff
@@ -11,11 +11,13 @@ import {
...
 import { getProtocolService } from "@services/Factory";
 import { useState } from "react";
+import { useNavigate } from "react-router-dom";
 import styled from "styled-components";
 
 const FCL_UNIT = BigInt(10 ** 12);
 
-export function SendTokens(props: { onFinish: () => void }) {
+export function SendTokens(props: { onFinish: string }) {
+    const navigate = useNavigate();
```
shelbyd, line 20: Run formatter.

data-host/library/src/popup/components/SendTokens.tsx
```diff
@@ -32,6 +34,7 @@ export function SendTokens(props: { onFinish: () => void }) {
       amount={amount}
       onChangeAmount={setAmount}
       onContinue={() => setPage("confirm")}
+      onCancel={() => navigate(props.onFinish)}
```
shelbyd, line 37: I don't like that this component knows about navigation. `onFinish` should be a callback and the calling component can navigate.

data-host/library/src/popup/components/SendTokens.tsx
```diff
@@ -68,11 +71,12 @@ function SpecifySend(props: {
   amount: bigint;
   onChangeAmount: (a: bigint) => void;
   onContinue: () => void;
+  onCancel: () => void;
 }) {
   const addressError = getAddressError(props.address);
   const validAmount = props.amount > BigInt(0);
   const isValid = addressError == null && validAmount;
-
+  const [loading] = useState(false);
```
shelbyd, line 79: Why are you using a state with no setting? AFAIK, this is equivalent to just `false`.

shelbyd (changes requested):

data-host/library/src/popup/components/About.tsx
nikoladimitroff: I'd assume that `<Text>` has attached CSS?

data-host/library/src/popup/components/SendTokens.tsx
melatron: It is not possible to initialise the navigate function in the Router where the SendTokens component is used. Maybe an alternative is to use Link component from the react-router library instead?

data-host/library/src/popup/components/DataScreen/WebpageViews.tsx
melatron: There was a moment in which the facts were wrongly sent. Thats when I added the check and left it there. I can remove it.

data-host/library/src/lib/background.ts
melatron: `const self = this` is needed because the lambda is called in a different context. I will try if changing it from function to lambda would fix it.

data-host/chrome-extension/src/pages/Popup/index.tsx
melatron: It is currently working.

data-host/library/src/popup/components/SendTokens.tsx
melatron: Formatter should always be applied, otherwise the CI would fail.

data-host/library/src/lib/background.ts
```diff
@@ -1,5 +1,32 @@
...
+    });
+
+    await getMintingRegistrar().maybeTryRegister();
+  }
+  addListeners() {
+    //TODO(melatron): Create more generic method that handles all type of messages (for different facts)
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
+    chrome.runtime.onMessage.addListener((request, _sender, sendResponse) => {
+      self.addWebpage(request.content.hostname);
```
shelbyd, line 28: await?

data-host/library/src/popup/components/About.tsx
shelbyd: Probably, but my point is we want to use margins or something to space things. I'm not aware of any legitimate use of `<br />`, except for doing spacing that should be done with CSS.

data-host/library/src/popup/components/DataScreen/index.tsx
```diff
@@ -105,9 +105,7 @@ function DataScreen() {
       <Address address={address} />
       <Button
         onClick={() =>
-          activityStack.push(
-            <SendTokens onFinish={() => activityStack.pop()} />,
-          )
+            navigate(RoutesPaths.SEND)
```
shelbyd, line 108: Formatting should put this one the same line.

`onClick={() => navigate(RoutesPaths.SEND)}`

data-host/library/src/popup/components/SendTokens.tsx
shelbyd: We discussed this and can use whatever return-to-previous navigation exists. I still prefer this component not knowing about navigation and just having a callback when it is finished.

data-host/library/src/popup/components/SweepTokens.tsx
```diff
@@ -11,19 +11,20 @@ import {
...
 import { useLoadedState } from "@utils/ReactHooks";
 import { useState } from "react";
+import { useNavigate } from "react-router-dom";
 import styled from "styled-components";
-
 const FCL_UNIT = BigInt(10 ** 12);
 
-export function SweepTokens(props: { onFinish: () => void }) {
+export function SweepTokens(props: { onFinish: string }) {
+    const navigate = useNavigate();
```
shelbyd, line 19: Why does this line have 4 spaces and the line below have 2?

shelbyd (changes requested):

data-host/library/src/popup/services/ProtocolOptIn.ts
```diff
@@ -74,6 +73,7 @@ export class ProtocolOptIn {
 
   private async tryRegisterIdentity(onMissingLiveness?: () => Promise<void>) {
     try {
+      console.log('--------------------------------------------');
```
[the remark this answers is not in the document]
melatron: Removed in the Jest PR, but I will remove it also here.

PR #104, opened 2022-01-12
Remove Ava and add Jest tests.

melatron:


data-host/library/src/popup/services/ProtocolOptIn.test.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    it('sets liveness on instances', async () => {
+      const { protocol, optIn } = graph();
+
+      await optIn.optIn('some mnemonic');
+
+      await optIn.postOptInLiveness();
+
+      expect(protocol.ensureIdentityRegistered).toHaveBeenCalled();
+    });
```
shelbyd, line 109: This test is exactly the same as the previous test. Also, it doesn't match the description. I'd just delete this one.

data-host/library/src/popup/services/StorageService.test.ts
```diff
@@ -0,0 +1,310 @@
...
+import {
+  StorageServiceClear as ERROR_CLEAR,
+  StorageServiceGetItem as ERROR_GET_ITEM,
+  StorageServiceHasItem as ERROR_HAS_ITEM,
+  StorageServiceRemoveItem as ERROR_REMOVE_ITEM,
+  StorageServiceSetItem as ERROR_SET_ITEM,
+} from '@services/StorageService';
+import { chrome } from 'jest-chrome';
+
+console.error = jest.fn();
```
shelbyd, line 11: We shouldn't override console.error. If we do, we should do it just for one test and then reset back to what it was.

shelbyd (approved):

PR #109, opened 2022-02-03
Parse json to sier

melatron:


support/sier-codec/src/lib.rs
```diff
@@ -48,13 +50,79 @@ impl Parser {
...
+                    list.push(val);
+                }
+                Ok((Type::List(Box::new(arr_type)), Value::List(list)))
+            }
+            SerdeValue::Object(inner_json_obj) => {
+                let (_object, def) = self.transform_serde_obj(inner_json_obj)?;
+                Ok((
+                    Type::Struct(def),
+                    Value::Unit,
+                    //TODO(melatron): Figure how to handle the lifetime of object - Value::Struct(object),
```
shelbyd, line 84: IIRC, our Value type has the lifetime parameter of the struct definitions we're working with. So, the 'a in the function definition should probably be the lifetime of the &self (shouldn't need to be mutable). I think changing `&mut self` to `&'a mut self` will let things compile (if they aren't already).

support/sier-codec/src/lib.rs
```diff
@@ -48,13 +50,79 @@ impl Parser {
...
-impl Default for Parser {
-    fn default() -> Self {
-        Parser {
-            structs: HashMap::new(),
+    fn transform_serde_value<'a>(
+        &mut self,
+        value: &SerdeValue,
+    ) -> Result<(Type, Value<'a>), Error<'a>> {
+        match value {
+            SerdeValue::Null => Ok((Type::Unit, Value::Unit)),
```
shelbyd, line 59: I don't think we want to parse null as (). I'd have to think through use cases to be sure, but it doesn't seem right.

support/sier-codec/src/lib.rs
```diff
@@ -48,13 +50,79 @@ impl Parser {
...
+            SerdeValue::Bool(_bool) => Ok((Type::Unit, Value::Unit)),
```
shelbyd, line 60: Don't we have a bool type?

support/sier-codec/src/lib.rs
```diff
@@ -48,13 +50,79 @@ impl Parser {
...
+            SerdeValue::Number(number) => {
+                if let Some(n) = number.as_u64() {
+                    Ok((Type::U64, Value::U64(n)))
+                } else {
+                    Ok((Type::Unit, Value::Unit))
```
shelbyd, line 65: We should error or return float? I think it's very confusing to return Unit when an error happens. While developing, you can leave `todo!()` or `unimplemented!("<details>")` for cases you haven't covered yet.

support/sier-codec/src/lib.rs
melatron: We don't, but I guess I will implement it in this PR.

support/sier-codec/src/lib.rs
melatron: What should we parse it as? Maybe we need a Null type implementation?

support/sier-codec/src/lib.rs
```diff
@@ -48,14 +50,76 @@ impl Parser {
     pub fn struct_def(&self, name: &str) -> Option<&Arc<StructDef>> {
         self.structs.values().find(|s| s.type_name() == name)
     }
-}
 
-impl Default for Parser {
-    fn default() -> Self {
-        Parser {
-            structs: HashMap::new(),
+    fn transform_serde_value<'a>(
```
shelbyd, line 54: As discussed in our VC, these should be free functions (without a self parameter).

support/sier-codec/src/lib.rs
```diff
@@ -48,14 +50,76 @@ impl Parser {
...
+        &'a self,
+        value: &SerdeValue,
+        type_: &'a Type,
+    ) -> Result<Value<'a>, Error<'a>> {
+        match value {
+            SerdeValue::Null => Ok(Value::Unit),
+            //TODO(melatron): Implement a boolean type for sier
+            SerdeValue::Bool(_bool) => Ok(Value::Unit),
```
shelbyd, line 62: Keep the arms of this match as `unimplemented!()` for now where we don't have a good implementation. Null, Bool, Number<float>.

support/sier-codec/src/lib.rs
```diff
@@ -48,14 +50,76 @@ impl Parser {
...
+            values.push(sier_value);
+        }
+
+        Ok(Object::new(def.as_ref(), values))
+    }
+
+    pub fn parse_json<'a>(
+        &'a self,
+        file_json_contents: &str,
+        def: &'a Arc<StructDef>,
```
shelbyd, line 115: Should be fine as a raw ref, `&'a StructDef`.

support/sier-codec/src/lib.rs
```diff
@@ -70,6 +134,7 @@ pub enum Error<'i> {
     TooFewBytes,
     TooManyBytes,
     InvalidUtf8(std::str::Utf8Error),
+    InvalidJson,
```
shelbyd, line 137: We should include some extra information about what the error is. Otherwise it will be hard for the users to know what's wrong. Perhaps we should look into anyhow or error_chain or whatever new crate people are using for errors.

support/sier-codec/src/lib.rs
shelbyd: Also in a separate file. json.rs or something similar.

support/sier-codec/src/lib.rs
melatron: I think we should start using `thiserror` in the whole lib. This means I should refactor the error handling everywhere. Maybe it would be better to do it in another PR? For this one I added one more type of error.

shelbyd: Happy to use whatever error library is the current standard in Rust (they change all the time). Just make sure it doesn't have `std` requirements.

We don't want to have `MismatchTypeFromJson` long term, and removing a variant of an enum is a breaking change. Let's just keep `InvalidJson` for now, and create Json specific errors in that module once we integrate/use `thiserror`.

support/sier-codec/tests/integrated.rs
```diff
@@ -164,3 +164,55 @@ fn struct_() {
...
+"#;
+
+const JSON: &'static str = r#"
+{
+    "sierType": "Foo",
+    "bar": 42,
+    "baz": "abc",
+    "qux": [4, 2],
+    "corge": {
+        "sierType": "Corge",
```
shelbyd, line 190: We shouldn't need a marker for the SIER type. We should be able to know the type from the struct defs.

support/sier-codec/tests/integrated.rs
```diff
@@ -164,3 +164,55 @@ fn struct_() {
...
+const JSON_STRUCT_DEF: &'static str = r#"
+struct Foo {
+    bar :u64;
+    baz :string;
+    qux :List<u64>;
+    corge :Corge;
+}
+"#;
+
+const JSON_STRUCT_DEF2: &'static str = r#"
```
shelbyd, line 177: Do we need this as a separate string? I thought we supported multiple structs per file.

support/sier-codec/src/lib.rs
```diff
@@ -48,13 +52,17 @@ impl Parser {
...
-        Parser {
-            structs: HashMap::new(),
-        }
+    pub fn parse_json<'a>(
+        &'a self,
+        file_json_contents: &str,
+        def: &'a Arc<StructDef>,
+    ) -> Result<Object, Error<'a>> {
+        let json: SerdeValue =
+            serde_json::from_str(file_json_contents).or(Err(Error::InvalidJson))?;
```
shelbyd, line 62: `or(Err(...))` should be `map_err` I think. I'm surprised clippy doesn't say something about it.

shelbyd: Actually, I don't know if clippy runs against this directory...

support/sier-codec/src/lib.rs
```diff
@@ -48,13 +52,17 @@ impl Parser {
...
-}
 
-impl Default for Parser {
-    fn default() -> Self {
-        Parser {
-            structs: HashMap::new(),
-        }
+    pub fn parse_json<'a>(
+        &'a self,
+        file_json_contents: &str,
```
shelbyd, line 58: I think the name should be `json_str` since it doesn't necessarily need to be a file. And parameter names are included in documentation, so should be very high quality.

support/sier-codec/src/json.rs
```diff
@@ -0,0 +1,69 @@
+use crate::{Error, Object, StructDef, Type, Value};
+use serde_json::Value as SerdeValue;
+use std::sync::Arc;
+
+pub fn transform_serde_value<'a>(
```
shelbyd, line 5: "transform" implies to me that we're mutating the object. I would still use the word "parse", so `parse_serde_value`. Same below.

support/sier-codec/src/json.rs
```diff
@@ -0,0 +1,69 @@
...
+    match value {
+        SerdeValue::Null => unimplemented!(),
+        SerdeValue::Bool(_bool) => unimplemented!(),
+        SerdeValue::Number(number) => {
+            if type_ == &Type::U64 {
+                if let Some(n) = number.as_u64() {
+                    Ok(Value::U64(n))
+                } else {
+                    unimplemented!()
+                }
```
shelbyd, line 18: This can be `number.as_u64().ok_or_else(|| unimplemented!())`.

support/sier-codec/src/json.rs
```diff
@@ -0,0 +1,69 @@
...
+
+pub fn transform_serde_value<'a>(
+    value: &SerdeValue,
+    type_: &'a Type,
+) -> Result<Value<'a>, Error<'a>> {
+    match value {
+        SerdeValue::Null => unimplemented!(),
+        SerdeValue::Bool(_bool) => unimplemented!(),
+        SerdeValue::Number(number) => {
+            if type_ == &Type::U64 {
```
shelbyd, line 13: I am a big fan of early returns. For this case, I'd change the code to something like:

```rs
if *type_ != Type::U64 {
  return Err(...);
}

let n = number.as_u64().ok_or_else(|| unimplemented!())?;
Ok(Value::U64(n))
```

In general, it makes it very easy to tell how you're handling incorrect cases. It also keeps the code flat and more readable.

support/sier-codec/src/json.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+pub fn transform_serde_obj<'a>(
+    json_obj: &serde_json::Map<String, SerdeValue>,
+    def: &'a Arc<StructDef>,
+) -> Result<Object<'a>, Error<'a>> {
+    let mut values = Vec::with_capacity(json_obj.len());
+    for field_def in def.fields().iter() {
+        let sier_value = transform_serde_value(
+            json_obj.get(field_def.name()).unwrap_or(&SerdeValue::Null),
```
shelbyd, line 62: I don't think we want to default to null. I think we want to error if we're missing a field. The JSON can explicitly include a null if it makes sense for the specific struct.

shelbyd (changes requested):

support/sier-codec/src/schema.rs
```diff
@@ -18,7 +18,7 @@ impl StructDef {
     }
 
     pub fn id(&self) -> [u8; 8] {
-        let mut hasher = Blake2b::new();
+        let mut hasher = Blake2s256::new();
```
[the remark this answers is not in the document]
melatron: With the newest version of the crate we should use `Blake2s256` or `Blake2b512` and I think that `Blake2b` default was `Blake2s256`.

support/sier-codec/src/lib.rs
melatron: Clippy runs and isn't finding any problems. Isn't map_err accepting a Function? Here I just need to transform the error, I am not sure which is the best way?

support/sier-codec/src/json.rs
melatron: But we need to wrap the `u64` into `Value::U64` and the ok_or_else returns a `Result`. Can you please provide an example, because I don't it would look cleaner this way.

support/sier-codec/src/json.rs
melatron: How should this be done for the `SerdeValue::Array` and `SerdeValue::Struct`? For the String and Number I agree.

support/sier-codec/src/json.rs
melatron: How should this be done for the SerdeValue::Array and SerdeValue::Struct? For the String and Number I agree.

support/sier-codec/src/json.rs
```diff
@@ -0,0 +1,63 @@
...
+            Ok(Value::String(s.to_string()))
+        }
+        SerdeValue::Array(vec) => {
+            if let Type::List(arr_type) = type_ {
+                let mut list = vec![];
+                for arr_value in vec {
+                    let val = parse_serde_value(arr_value, arr_type)?;
+                    list.push(val);
+                }
+                Ok(Value::List(list))
```
shelbyd, line 31: You can collect an `Iterator<Result<T, E>>` into a `Result<Vec<T>, E>`. So might be nicer as:

```rs
let list = vec
    .into_iter()
    .map(|val| parse_serde_value(val, arr_type))
    .collect::<Result<_, _>>()?;
Ok(Value::List(list))
```

support/sier-codec/src/schema.rs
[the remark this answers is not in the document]
shelbyd: Is there any relevant tradeoff between 2s256 and 2b512?

shelbyd (approved):

support/sier-codec/src/schema.rs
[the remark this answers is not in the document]
melatron: From what I found - "BLAKE2b and BLAKE2s are designed to be efficient on a single CPU core (BLAKE2b is more efficient on 64-bit CPUs and BLAKE2s is more efficient on 8-bit, 16-bit, or 32-bit CPUs)"

Changed it to Blake2b512

PR #121, opened 2022-03-01
Shelbyd+melatron/send funds

shelbyd:


scripts/send_funds.js
```diff
@@ -1,30 +1,60 @@
...
-const args = require('minimist')(process.argv.slice(2));
+const args = require("minimist")(process.argv.slice(2));
 
 // Usage:
 //   node send_funds.js \
 //     --amounts send_funds_amounts_example.csv \
-//     --network wss://nodes.testnet.fractalprotocol.com
+//     --out ./sent_txns.csv \
+//     --network wss://nodes.testnet.fractalprotocol.com \
+//     --send-at-once 256
```
melatron, line 16: If we make the comment with /**/ it would be easy to copy the command.

melatron (approved):

scripts/send_funds.js
shelbyd: True, I don't want to do that right now. I'll remember it for next time I'm working on this file.

PR #124, opened 2022-03-22
Start using thiserror crate for sier codec.

melatron:


shelbyd (approved):
I tweaked the error messages a bit to make them more succinct and consistent. We can iterate on them if you'd like, but please let me review if you do change them.

PR #125, opened 2022-03-22
Implement boolean type for sier-codec

melatron:


support/sier-codec/src/json.rs
```diff
@@ -5,7 +5,13 @@ use std::sync::Arc;
 pub fn parse_serde_value<'a>(value: &SerdeValue, type_: &'a Type) -> Result<Value<'a>, Error<'a>> {
     match value {
         SerdeValue::Null => unimplemented!(),
-        SerdeValue::Bool(_bool) => unimplemented!(),
+        SerdeValue::Bool(bool) => {
```
shelbyd, line 8: Since `bool` is a keyword in Rust, it's weird to use as a variable name. Just `b` would be clear enough from context. Alternatively `v` for value.

support/sier-codec/src/object.rs
```diff
@@ -151,6 +165,13 @@ impl<'s> Value<'s> {
     pub fn serialize(&self) -> Vec<u8> {
         match self {
             Value::Unit => Vec::new(),
+            Value::Bool(v) => {
+                if *v {
+                    Vec::from(1u8.to_le_bytes())
+                } else {
+                    Vec::from(0u8.to_le_bytes())
+                }
```
shelbyd, line 173: You're duplicating the Vec::from and .to_le_bytes. I would do something like the following.

```rust
let byte: u8 = if *v { 1 } else { 0 };
Vec::from(byte.to_le_bytes())
```

support/sier-codec/src/schema.rs
```diff
@@ -146,6 +148,18 @@ impl Type {
...
         match self {
             Type::Unit => Ok((bytes, Value::Unit)),
+            Type::Bool => {
+                let (b, n) = complete::le_u8(bytes).map_err(Error::ValueParsing)?;
+                let value = match n {
+                    0 => Ok(Value::Bool(false)),
+                    1 => Ok(Value::Bool(true)),
+                    _ => Err(Error::ValueParsing(nom::Err::Error(
+                        nom::error::make_error(bytes, nom::error::ErrorKind::IsNot),
+                    ))),
```
shelbyd, line 158: This seems much more complicated than it should be. We could make a new variant of our Error type. It's also not the worst to keep it how it is now.

shelbyd, line 158: I'd have the match statement just be true/false and the error case explicitly return.

```rust
let value = match n {
  0 => false,
  1 => true,
  _ => return Err(...),
};
Ok((b, Value::Bool(value)))
```

shelbyd (approved):

PR #137, opened 2022-04-02
Begin implementation of Cross Chain Object Storage

shelbyd:
This is starting to get too big. I can split this up into fewer PRs if this is too much for one.

### DIFF
--- a/object_storage/src/fractal_store.rs
+++ b/object_storage/src/fractal_store.rs
@@ -1,0 +1,119 @@
+use crate::*;
+
+#[mockall_double::double]
+use crate::kv_set::KvSet;
+
+pub struct FractalStore<D> {
+    db: D,
+    kv_set: KvSet,
+}
+
+impl<D: Database + 'static> FractalStore<D> {
+    pub fn new(db: D) -> Self {
+        Self::new_with_deps(db, KvSet::new())
+    }
+
+    fn new_with_deps(db: D, kv_set: KvSet) -> Self {
+        FractalStore { db, kv_set }
+    }
+
+    pub fn init(&mut self, id: &[u8], value: &[u8]) -> Result<(), Error<D::Error>> {
+        if self.kv_set.contains(id, &self.db)? {
+            return Err(Error::IdExists);
+        }
+
+        self.db.store(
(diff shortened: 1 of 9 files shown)


object_storage/tests/integrated.rs
```diff
@@ -0,0 +1,21 @@
+use fractal_object_storage::{test::InMemoryDb, FractalStore, Given, ProofChecker, Proposition};
+
+#[test]
+fn proving_simple_value() {
```
melatron, line 4: Maybe we can add more integrated tests for cases like - multiple same proofs, removing proofs, etc.

object_storage/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+            Ok(p) => p,
+            Err(_) => return false,
+        };
+
+        self.verify_proposition(proposition, proof)
+    }
+
+    fn verify_proposition(&self, prop: Proposition, proof: Proof) -> bool {
+        match (prop, proof) {
+            (Proposition::ObjectIsValue(object_id, value), p) => {
```
melatron, line 57: Currently, proof can only be Empty, what are the potential enum properties and is this case ever going to be hit?

object_storage/src/fractal_store.rs
```diff
@@ -0,0 +1,265 @@
...
+
+        Ok(())
+    }
+
+    pub fn root_hash(&self) -> Result<Hash, D::Error> {
+        let mut merkle_tree = MerkleTree::new();
+
+        let mut object_ids = self.objects.iter_lexicographic();
+        while let Some(object_id) = object_ids.next()? {
+            let data = self.object_data(&object_id)?.unwrap();
```
melatron, line 46: Add expect with info why this would never alert.

object_storage/src/fractal_store.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+impl<D> Clone for PrefixedHandle<D> {
+    fn clone(&self) -> Self {
+        PrefixedHandle {
+            prefix: self.prefix.clone(),
+            handle: self.handle.clone(),
+        }
+    }
+}
```
melatron, line 188: Why not just use `#[derive(Clone)]`?

melatron (approved):

PR #140, opened 2022-04-07
Sier codec wasm binding.

melatron:


support/sier-codec-js/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    let json = js_sys::JSON::stringify(&js_object).unwrap_or_else(|e| {
+        wasm_bindgen::throw_val(e);
+    });
+    let json = json
+        .as_string()
+        .expect("Parsed as string from JSON::stringify");
+
+    let mut parser = Parser::default();
+    parser.add_file_defs(file_defs).unwrap_or_else(|e| {
+        wasm_bindgen::throw_str(format!("Wrong file definitions provided {0}", e).as_str());
```
shelbyd, line 20: Errors here aren't necessarily "Wrong file definitions". I would just pass through the error (possibly stringified).

support/sier-codec-js/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    let def = parser.struct_def(struct_def).unwrap_or_else(|| {
+        wasm_bindgen::throw_str("Wrong struct_def provided");
```
shelbyd, line 24: This error would be something like `Could not find struct {0} in provided definitions`.

support/sier-codec-js/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    let obj = parser.json_str(json.as_str(), def).unwrap_or_else(|e| {
+        wasm_bindgen::throw_str(format!("Wrong JSON provided - {0}", e).as_str());
```
shelbyd, line 28: Again, just pass through the error.

shelbyd (changes requested):

support/sier-codec-js/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    let json = js_sys::JSON::stringify(&js_object).unwrap_or_else(|e| {
+        wasm_bindgen::throw_val(e);
+    });
+    let json = json
+        .as_string()
+        .expect("Parsed as string from JSON::stringify");
+
+    let mut parser = Parser::default();
+    parser.add_file_defs(file_defs).unwrap_or_else(|e| {
+        wasm_bindgen::throw_str(e.to_string().as_str());
```
shelbyd, line 20: Instead of `throw_str(s.as_str())` you should be able to `throw_str(&s)`. Same elsewhere.

support/sier-codec-js/src/lib.rs
```diff
@@ -0,0 +1,54 @@
+use sier_codec::Parser;
+use wasm_bindgen::prelude::*;
+//TODO(melatron): Create a JS Class that holds the file_defs inside and
+// serialize/deserialize are methods of this class.
+#[wasm_bindgen]
+pub fn serialize(js_object: JsValue, file_defs: &str, struct_def: &str) -> js_sys::Uint8Array {
+    if !js_object.is_object() {
+        wasm_bindgen::throw_str("Provided argument is not a JS Object.");
```
shelbyd, line 8: You might be able to have these functions return `Result<T, JsValue>` instead of explicitly throwing everywhere. More idiomatic and probably easier to read/write. Might even be able to use the `?` operator.

support/sier-codec-js/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    let json = js_sys::JSON::stringify(&js_object).unwrap_or_else(|e| {
+        wasm_bindgen::throw_val(e);
+    });
```
shelbyd, line 13: This should be doable as:

```
let json = js_sys::JSON::stringify(&js_object)?;
```

shelbyd (changes requested):

PR #155, opened 2022-05-19
Add pallet allowing users to stake and receive rewards

shelbyd:
Re: #156 

Some questions @nikoladimitroff

1. Should we have a minimum stake amount? Power users may make frequent staking TXNs, but I don't want to gatekeep poorer users.
2. What should the staking lock period be? I set it to 30 days for now, but we can change it.

Small TODOs before submitting:

- [x] Minimum stake amount
- [x] Stake for multiple durations.
- [X] Investigate using [ReservableCurrency](https://docs.rs/frame-support/latest/frame_support/traits/trait.ReservableCurrency.html) instead of our own holding account.
- [x] Increment amount instead of replacing if multiple stakes for a user in the same block.

### DIFF
--- a/blockchain/pallets/staking/src/lib.rs
+++ b/blockchain/pallets/staking/src/lib.rs
@@ -53,9 +53,7 @@ pub mod pallet {
     #[pallet::metadata(BalanceOf<T> = "Balance")]
     #[pallet::generate_deposit(pub fn deposit_event)]
     pub enum Event<T: Config> {
-        Distribution {
-            amount: BalanceOf<T>,
-        }
+        Distribution { amount: BalanceOf<T> },
     }
 
     #[pallet::error]
@@ -67,7 +65,7 @@ pub mod pallet {
 
     #[pallet::call]
     impl<T: Config> Pallet<T> {
-        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(0, 1))]
+        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(2, 2))]
         pub fn stake(
             origin: OriginFor<T>,
             #[pallet::compact] amount: BalanceOf<T>,
@@ -91,7 +89,7 @@ pub mod pallet {
             Ok(())
         }
 
-        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(0, 1))]
+        #[pallet::weight(10_000 + T::DbWeight::get().reads_writes(10, 10))]
         pub fn withdraw(
             origin: OriginFor<T>,
             #[pallet::compact] amount: BalanceOf<T>,
@@ -109,7 +107,7 @@ pub mod pallet {
                 .fold(BalanceOf::<T>::zero(), |acc, b| acc + b);
 
             if total_unlocked < amount {
-                return Err(Error::<T>::NotEnoughUnlockedStake)?;
+                return Err(Error::<T>::NotEnoughUnlockedStake.into());
             }
 
             let mut blocks = unlocked_balances.into_keys();
--- a/blockchain/pallets/staking/src/lib.rs
+++ b/blockchain/pallets/staking/src/lib.rs
@@ -137,7 +137,7 @@ pub mod pallet {
                 address,
                 <frame_system::Pallet<T>>::block_number() + lock_period,
                 |map| {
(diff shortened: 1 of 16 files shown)


nikoladimitroff:
@shelbyd 
1. the avg daily income right now is > 4 FCL so I think a minimal stake amount of 10 FCL is reasonable (~3 days worth of current data capture, ~5 days worth of future KYC + data capture)
2. Let's have a couple of locking periods which increase the number of shares you get from the rewards (just like we wanted the holder program to reward people with more coindays). Let's have 30d, 60d, 90d as options for now

shelbyd:
@nikoladimitroff Do we want to automatically withdraw the staked amount (plus new tokens) when the lock period expires, or allow it to stay staked, but unlocked, until the user chooses to withdraw?

With manual withdrawals, the user could have a significant balance currently unlocked and be receiving a large share of rewards. For example, if they locked for the maximum duration and haven't withdrawn.

With automatic withdrawals, users need to re-stake tokens and actively maintain this locked stake.

There are some more benefits / drawbacks, but I wanted to start the conversation. We can also VC about it with Julio and/or Antoni if we think that's necessary.

nikoladimitroff:
> @nikoladimitroff Do we want to automatically withdraw the staked amount (plus new tokens) when the lock period expires, or allow it to stay staked, but unlocked, until the user chooses to withdraw?
> 
> With manual withdrawals, the user could have a significant balance currently unlocked and be receiving a large share of rewards. For example, if they locked for the maximum duration and haven't withdrawn. Additionally, users will need to pay a TXN fee to withdraw their stake.
> 
> With automatic withdrawals, users need to re-stake tokens and actively maintain this locked stake.
> 
> There are some more benefits / drawbacks, but I wanted to start the conversation. We can also VC about it with Julio and/or Antoni if we think that's necessary.

@shelbyd I propose a third one - similar to how bank deposit work.
1. You lock some currency for a certain amount of time
2. After that time expires, you get access to your funds + the interest
3. You could choose to re-deposit. You could choose to withdraw. You could just to do nothing and the funds just stay there without earning anything further.

This approach:
- forces people to actively manage their staked funds (which increases engagement)
- doesn't reward people who haven't manually withdrawn their funds as in your first proposal
- since people are incentivized to either restake or withdraw, this gives us a better understanding of the amount of FCL currently locked and the liquid amount
- allows users to restake without paying a TXN fee

shelbyd:
This seems effectively identical to automatic withdrawals, but with more explanatory and implementation complexity. I think that we should go with automatic withdrawals for all the reasons listed.

Note: Users would still need to pay a TXN fee to restake/withdraw. Users need to pay TXN fee for any user-initiated on-chain activity.

blockchain/pallets/staking/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+        fn on_finalize(block_number: BlockNumberFor<T>) {
+            if block_number % T::DistributeEveryNBlocks::get() != 0u32.into() {
+                return;
+            }
+
+            let to_distribute = T::Currency::free_balance(&T::DistributionSource::get());
+            let total_staked = TotalCoinShares::<T>::get();
+
+            let mut distributed = BalanceOf::<T>::zero();
+            let mut distributed_shares = BalanceOf::<T>::zero();
```
nikoladimitroff, line 174: put a comment on top explaining the formula

````
// The formula is roughly
// user_shares = % of total staked FCL + length of time
// earned_fcl = total_fcl_to_distribute * user_shares / total_shares
```

nikoladimitroff (approved):

blockchain/pallets/staking/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+                T::Currency::transfer(
+                    &T::DistributionSource::get(),
+                    &address,
+                    amount,
+                    ExistenceRequirement::AllowDeath,
+                )
+                .expect("distributing based on balance");
+                T::Currency::reserve(&address, amount).expect("just deposited to this account");
+
+                if block <= block_number {
```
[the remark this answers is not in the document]
shelbyd: Done.

blockchain/pallets/staking/src/lib.rs
shelbyd: Refactored a lot. Hopefully the logic is clearer now.

PR #167, opened 2022-10-13
WIP: UI implementation

juliosantos:


explorer/src/swap_chains/drive.rs
```diff
@@ -35,11 +35,43 @@ fn drive_sending(
...
+                        swap_id TEXT PRIMARY KEY NOT NULL
+                    )",
+                    &[],
+                )?;
+                return Ok(do_insert(pg)?);
+            }
+        }
+
+        anyhow::bail!(e);
+    }
```
pkoch, line 74: I'd do this migration (adding the table) out of band (i.e., not here, inside normal operation code, but instead on an interactive session).

juliosantos: This is the same pattern we have in storage.rs. I don't disagree with the comment, but unless you're pointing to anything other than us not following the good 12-factor practices, I think it's out of scope

pkoch (commented):
This is big, and not passing tests yet. Can you guide on what you're looking from me as a reviewer?

Here's some preliminary stuff, meanwhile.

juliosantos:
I made this PR "not a draft", and since you were a reviewer before GitHub pinged you automatically. Thanks for taking the time, nothing I need here!

> not passing tests
It's warnings only, the check fails because circle's running `cargo test --release` with `RUSTFLAGS="-D warnings"`

explorer/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "dependencies": {
```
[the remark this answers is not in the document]
juliosantos: we've `yarn.lock` in VC

